PR #1, opened 2025-09-04
Add support to download all referenced packages from snapshot mirror

fmoessbauer:
For license clearing it is important to get the source of a package (both source and binary packages). For that, we add the subcommand "download" to download all referenced debian packages to a local directory. To avoid time consuming downloads on changes, we check upfront if we already have the package and only download if not.

### DIFF
--- a/src/debsbom/download/__init__.py
+++ b/src/debsbom/download/__init__.py
@@ -1,0 +1,6 @@
+# Copyright (C) 2025 Siemens
+#
+# SPDX-License-Identifier: MIT
+
+from .download import PackageResolver
+from .download import PackageDownloader
--- a/src/debsbom/download/__init__.py
+++ b/src/debsbom/download/__init__.py
@@ -2,5 +2,5 @@
 #
 # SPDX-License-Identifier: MIT
 
-from .download import PackageResolver
+from .download import PackageResolver, PackageResolverCache, PersistentResolverCache
 from .download import PackageDownloader
--- a/src/debsbom/snapshot/client.py
+++ b/src/debsbom/snapshot/client.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2025 Siemens
+#
+# SPDX-License-Identifier: MIT
+
+from dataclasses import dataclass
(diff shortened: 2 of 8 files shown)


src/debsbom/snapshot/client.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2025 Siemens
+#
+# SPDX-License-Identifier: MIT
+
+from dataclasses import dataclass
+from typing import Generator, Type
+import requests
```
Urist-McGit, line 9: You are using the requests library here and have not added it to the dependencies.

fmoessbauer: Shall we add it as optional dependency and only load the infrastructure when calling the download command, or shall I just require it?

Urist-McGit: I think requests is ubiquitous enough we can just add it

src/debsbom/snapshot/client.py
```diff
@@ -0,0 +1,192 @@
...
+                raise NotFoundOnSnapshot()
+            data = r.json()
+        except RequestException as e:
+            raise SnapshotDataLakeError(e)
+
+        fileinfo = data.get("fileinfo")
+        for s in data.get("result", []):
+            hash = s["hash"]
+            # TODO: this might be ambiguous if the same file is uploaded
+            # under different names
```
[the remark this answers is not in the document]
Urist-McGit: I think we can ignore that for now

src/debsbom/download/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (C) 2025 Siemens
+#
+# SPDX-License-Identifier: MIT
+
+from .download import PackageResolver
```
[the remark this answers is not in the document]
fmoessbauer: Done!

fmoessbauer:
@Urist-McGit I just rebased, improved the remaining "TODO" messages, updated the README and fixed a typo in the name of the `SnapshotNotFoundError` exception. By that, this one should be ready to go.

PR #5, opened 2025-09-08
fix(dpkg): handle version type conversion

fmoessbauer:
The SourcePackage and BinaryPackage version must be converted to a deb822 version type (even if it is provided as string) to later be able to access parts of the version (e.g. the epoch / source version / debian revision).

We fix this by converting the version in the respective constructor. This avoid having to fix all the callers and also avoid exposing the dep822 type to all downstream users. When passing the version of a package to a downstream component, we further need to ensure we convert it to its string representation first.

Fixes: 958c97d ("Initial commit")


Urist-McGit:
LGTM. I just noticed our CI did not run properly for PRs. Can you please rebase so the pipelines actually run?

PR #6, opened 2025-09-08
feat: add source-merge subcommand

fmoessbauer:
We add the merge subcommand which runs against a download directory and creates combined archive files, which can be used as input to license clearing tools that only support a single archive per component.

### DIFF
--- a/src/debsbom/download/download.py
+++ b/src/debsbom/download/download.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 from abc import abstractmethod
+from collections import namedtuple
 import dataclasses
 from functools import reduce
 import hashlib
@@ -154,6 +155,8 @@ class PackageResolver:
 
 
 class PackageDownloader:
+    StatisticsType = namedtuple("statistics", "files bytes cfiles cbytes")
+
     def __init__(
         self, outdir: Path | str = "downloads", session: requests.Session = requests.Session()
     ):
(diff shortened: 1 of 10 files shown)


Urist-McGit:
`merge` makes it sound like we merge SBOMs, a very common thing to do with them.

fmoessbauer:
> `merge` makes it sound like we merge SBOMs, a very common thing to do with them. Maybe something like `archive` instead?

Which name do you propose instead for the command?

src/debsbom/download/merger.py
```diff
@@ -0,0 +1,146 @@
...
+    # fmt: off
+    NONE  = _CmdAndExt("cat",   [],     [],                 "")
+    BZIP2 = _CmdAndExt("bzip2", ["-q"], ["-q", "-d", "-c"], ".bz2")
+    GZIP  = _CmdAndExt("gzip",  ["-q"], ["-q", "-d", "-c"], ".gz")
+    XZ    = _CmdAndExt("xz",    ["-q"], ["-q", "-d", "-c"], ".xz")
+    ZSTD  = _CmdAndExt("zstd",  ["-q"], ["-q", "-d", "-c"], ".zst")
+    # fmt: on
+
+    @staticmethod
+    def from_tool(tool: str):
```
Urist-McGit, line 37: Again please use a type hint.

src/debsbom/download/merger.py
```diff
@@ -0,0 +1,146 @@
...
+        if not tool:
+            return Compression.NONE
+        comp = [c for c in Compression.formats() if c.tool == tool]
+        if comp:
+            return comp[0]
+        raise RuntimeError(f"No handler for compression with {tool}")
+
+    @staticmethod
+    def from_ext(ext: str):
```
Urist-McGit, line 46: Type hint please.

src/debsbom/download/download.py
```diff
@@ -167,10 +167,12 @@ def register(self, files: list["sdlclient.RemoteFile"]):
 
     def stat(self):
```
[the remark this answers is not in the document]
fmoessbauer: While doing this, I also changed it to a named tuple.

PR #10, opened 2025-09-09
Refactor Package structure and BOM references

Urist-McGit:
Rework how we deal with Packages, Dependencies and BOM references to match the structure of the SBOM more closely.

### DIFF
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -2,54 +2,18 @@
 #
 # SPDX-License-Identifier: MIT
 
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from debian.deb822 import Packages, PkgRelation
 from debian.debian_support import Version
 from packageurl import PackageURL
 import re
 from typing import Iterator, List, Tuple, Type
-import functools
 
 from ..sbom import CDX_REF_PREFIX, SBOMType, SPDX_REF_PREFIX
 
 # SPDX IDs only allow alphanumeric, '.' and '-'
 SPDX_ID_RE = re.compile(r"[^A-Za-z0-9.\-]+")
-
-
-@dataclass(init=False)
-class SourcePackage:
-    """Representation of a Debian Source package."""
-
-    name: str
-    version: Version | None
-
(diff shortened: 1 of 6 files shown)


src/debsbom/sbom.py
```diff
@@ -33,3 +36,22 @@
...
+
+
+@dataclass
+class Reference:
+    """Generic reference in a SBOM."""
+
+    package_name: str
+    is_source: bool = False
+
+    def as_str(self, sbom_type: SBOMType) -> str:
```
fmoessbauer, line 48: Just use the string formatter `def __str__`.

src/debsbom/dpkg/package.py
```diff
@@ -86,15 +50,89 @@ def parse_depends_line(cls, line: str) -> List["Dependency"]:
...
+class Package(ABC):
+    """Base class for binary and source packages."""
 
     name: str
-    section: str
+    version: Version
+
+    def __init__(self, name: str, version: str | Version, maintainer: str):
+        self.name = name
+        self.version = Version(version)
```
fmoessbauer, line 50: The maintainer param is not used. Why do we need it here instead of just adding it to the constructor of derived classes?

src/debsbom/dpkg/package.py
```diff
@@ -86,15 +50,89 @@ def parse_depends_line(cls, line: str) -> List["Dependency"]:
...
+        self.maintainer = maintainer
+
+    def purl(self) -> PackageURL:
+        """Return the PURL of the package."""
+        return PackageURL.from_string(
+            "pkg:deb/debian/{}@{}?arch=source".format(self.name, self.version)
+        )
+
+    def bom_ref(self, sbom_type: SBOMType) -> str:
```
fmoessbauer: This should not be a member of the `SourcePackage`, but either a free function or a member of the sbom generator.

Urist-McGit: This is in the first commit of my series, where only the refactoring of the packages happen. In the second commit, where the actual reworking of the BOM references is done, it is removed. It even says `Outdated`

fmoessbauer (approved):

PR #17, opened 2025-09-12
docs: clearly describe scope of the tool in the README

fmoessbauer:


README.md
```diff
@@ -27,6 +27,32 @@ options:
...
+These SBOM outputs are designed to serve as reliable input for vulnerability management systems and license compliance checks.
+
+The tool provides auxiliary commands for package source retrieval. These enable users to:
+1. Retrieve packages from Debian's upstream repositories and report missing packages.
+2. Convert the multi-archive source packages into a single artifact (one archive per source package)
+
+At its core, this tool was designed to fulfill these SBOM generation requirements while maintaining:
+1. A minimal dependency footprint: avoid huge dependency graph of external software ecosystems (like Go or Rust)
+2. Strict focus on Debian-specific package formats
+3. Clear separation between binary packages and their corresponding source packages
```
gernot-h, line 48: What about adding:
3. Provide a thin layer on top of official CycloneDX and SPDX libraries

gernot-h: Perhaps the wording is not good enough, but if I understand it correctly, besides focusing on Debian, another reason why we invented this instead of contributing to distro2sbom was their decision to write and maintain their own cdx and SPDX library.

gernot-h (approved):
lgtm, thanks for taking care! From my side, ready for merging. @Urist-McGit, wdyt?

PR #19, opened 2025-09-12
feat: add apt cache repository parsing 

Urist-McGit:
Parse the apt cache lists in /var/lib/apt/lists to get more information
about involved source packages. This step is optional, as we only get
additional information for some source packages this way.

Note that there is a lot more information available in the apt cache
repositories and this is only a small portion of what is possible with it.


tests/root/apt-sources/var/lib/apt/lists/deb.debian.org_debian_dists_bookworm_main_source_Sources
```diff
@@ -0,0 +1,116 @@
...
+Standards-Version: 4.6.2
+Format: 3.0 (quilt)
+Files:
+ 6adb372f47c7b4b980d6c0bffae3f691 11692 binutils_2.40-2.dsc
+ 80b778f94b4a0250e3b2ee586ca449ff 25364048 binutils_2.40.orig.tar.xz
+ 9415c6640f763e28a3b73cf6b16051e8 102476 binutils_2.40-2.debian.tar.xz
+Vcs-Browser: https://salsa.debian.org/toolchain-team/binutils/tree/binutils-2.40
+Vcs-Git: https://salsa.debian.org/toolchain-team/binutils.git
+Checksums-Sha256:
+ cd75da7829d819189ba6154d408666373b307e222b393223804c4c4a7156f421 11692 binutils_2.40-2.dsc
```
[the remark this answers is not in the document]
fmoessbauer: Moved to separate MR.

fmoessbauer:
@Urist-McGit As discussed offline, I addressed my review comments and added a couple of performance improvements. Please review this as well - as I can't review my own changes for obvious reasons.

src/debsbom/apt/cache.py
```diff
@@ -85,7 +85,7 @@ def _parse_sources(
         try:
             with open(sources_file) as f:
                 logger.debug(f"Parsing apt cache sources: {sources_file}")
-                sources_raw = Sources.iter_paragraphs(f, use_apt_pkg=False)
+                sources_raw = Sources.iter_paragraphs(f, use_apt_pkg=True)
```
Urist-McGit, line 88: Not using apt-pkg is there for a reason unfortunately. The required dependency is [python-apt](https://pypi.org/project/python-apt/), which is unmaintained on PyPI. There is an up-to-date version distributed in Debian, which is installed per default in most systems. There is another occurrence when parsing the binary packages where we also deliberately do not use apt_pkg. If it is really such a speedup, maybe we should consider doing something about the `python-apt` situation in PyPI

Urist-McGit: The tests actually emit a warning too:
```
tests/test_generation.py::test_apt_source_pkg
  /opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/debian/deb822.py:723: UserWarning: Parsing of Deb822 data with python3-apt's apt_pkg was requested but this package is not importable. Is python3-apt installed?
    warnings.warn(msg)
```

fmoessbauer: That's indeed unfortunate. I added a commit to check for python `apt` upfront and enable this conditionally.

src/debsbom/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+# Optional transitive dependency of dep822 which is only distributed
+# via Debian (not pip). If not available but requested, the library
+# issues a warning which we want to avoid by checking upfront and
+# explicitly requesting the fallback mechanism.
+try:
+    import apt
+
+    HAS_PYTHON_APT = True
+except ImportError:
+    HAS_PYTHON_APT = False
```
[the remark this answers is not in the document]
fmoessbauer: That's tricky, as we are in the import statement of the module here. But we could emit a log message in the CLI module. How about that?

PR #24, opened 2025-09-17
feat: add support to generate SBOM according to standard-bom

fmoessbauer:
The standard-bom CycloneDX profiles/standard is owned by Siemens AG and specifies how an SBOM is to be layed out (and interpreted). We add support to add this information to the SBOM.

### DIFF
--- a/src/debsbom/cli.py
+++ b/src/debsbom/cli.py
@@ -14,6 +14,7 @@ from uuid import UUID
 from urllib.parse import urlparse
 from pathlib import Path
 
+from .sbom import BOM_Standard
 from .dpkg import package
 from .generate import Debsbom, SBOMType
 from . import HAS_PYTHON_APT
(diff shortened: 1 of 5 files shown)


src/debsbom/cli.py
```diff
@@ -127,6 +133,12 @@ def setup_parser(parser):
             help="vendor of debian distribution (debian or ubuntu)",
             default="debian",
         )
+        parser.add_argument(
+            "--standard",
+            choices=["unspecified", "standard-bom"],
+            help="generate SBOM according to this spec",
+            default="unspecified",
+        )
```
Urist-McGit, line 141: Finally our Siemens internal stuff finds its way here ;)

I am wondering though if the naming is abit misleading. Maybe something like `official` and `siemens-standard-bom` is better? If I do not know about Siemens Standard BOM I might think I want `standard-bom` instead of `unspecified`

fmoessbauer: > I am wondering though if the naming is abit misleading. Maybe something like official and siemens-standard-bom is better?

I don't know how well-established the term is. But at least other tooling like https://github.com/siemens/standard-bom-python also just uses the term `standard-bom`. @gernot-h 

> This also only matters for CDX, this should be mentioned here too. Now that I think of it, maybe we should instead name the argument --cdx-standard?

I prefer to keep it as a generic option to be able to later on use that standard on SPDX as well (or to add support for other standards which are also supported in SPDX).

PR #26, opened 2025-09-18
fix(merger): correctly handle filename if no debian revision

fmoessbauer:
If a source package has no debian revision, we added -None to the filename. This is incorrect, as we just need to leave out the revision in this case. We fix this by using the filename of the dsc as base and replace .dsc with .merged.tar.


Urist-McGit:
Nit: The fixes tag is missing

PR #28, opened 2025-09-18
ci: run tests on all supported Python versions

fmoessbauer:
As we already had fallouts on older Python versions, we better test them in the CI.


fmoessbauer:
The test did find correctly find the bug on 3.11, which is fixed in #27

PR #29, opened 2025-09-19
ci: add coverage reports

Urist-McGit:
Add a coverage report when we run the tests. In a second step, merge all
reports and make sure we score above 75%. This is not a hard goal, but
we have aroudn 80% right now, which is a good starting point.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ Repository = "https://github.com/siemens/debsbom.git"
 [project.optional-dependencies]
 dev = [
   "pytest>=6.0",
+  "pytest-cov>=6",
   "black",
   "beartype>=0.20",
   "debsbom[download]",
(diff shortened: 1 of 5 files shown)


Urist-McGit:
As discussed, I added the smoke test to the coverage. Only took me like 50 tries

Also I fixed the problem with CI running twice in all our PRs.

pyproject.toml
```diff
@@ -63,3 +64,15 @@ testpaths = [
...
+[tool.coverage.run]
+source = [
+  "debsbom"
+]
+
+[tool.coverage.paths]
+paths = [
+    "src/debsbom/",
+    "*/site-packages/debsbom/",
+    "*/runner/work/debsbom/debsbom/src/debsbom",
```
fmoessbauer, line 77: This is a path which probably is only valid on the GitHub CI. It would be better to not hardcode this in the description. Can't we set it in the execution step?

fmoessbauer (commented):
> but we have aroudn 80%

Nit: typo in commit 3.

pyproject.toml
fmoessbauer: Thanks for fixing this :smile:

PR #30, opened 2025-09-19
feat(generate): add source packages and built-using to CycloneDX format

fmoessbauer:
CycloneDX v1.6 does not have a proposed way of expressing the relation between a source package and a binary package in a machine readable way. Until this is fixed in the spec, we add the source packages similar to binary packages, where they can be distinguished by the PURL. We further add a dependency from the binary to the source package.

cc @gernot-h

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,15 +56,33 @@ At its core, this tool was designed to fulfill these SBOM generation requirement
 - Real-time vulnerability database integration
 - Signing and attestation of generated artifacts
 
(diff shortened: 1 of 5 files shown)


gernot-h:
I didn't review the source changes in detail, but the described approach looks good to me!

fmoessbauer:
While analyzing the generated CycloneDX output, I noticed that we have a bug in the linking of source and binary dependencies on CDX. Working on a fix.

PR #37, opened 2025-09-24
read, write and assert checksums

fmoessbauer:
As we recently introduced support to get package checksums from the apt-cache, we now use that information in the various SBOM processing steps to ensure the processed artifacts are identical with the artifacts used on the system that generated the SBOM.

Along that, there are minor improvements regarding the layout and handling of the downloads.

Note: This should go before #25

### DIFF
--- a/src/debsbom/download/spdx.py
+++ b/src/debsbom/download/spdx.py
@@ -6,12 +6,13 @@ from .download import PackageResolver
 from pathlib import Path
 from spdx_tools.spdx.parser.parse_anything import parse_file
 import spdx_tools.spdx.model.package as spdx_package
+import spdx_tools.spdx.model.document as spdx_document
 
 
 class SpdxPackageResolver(PackageResolver):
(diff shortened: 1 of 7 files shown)


src/debsbom/download/spdx.py
```diff
@@ -14,24 +17,26 @@ def __init__(self, document: spdx_document.Document):
...
+    def debian_pkgs(self) -> Iterable[Package]:
         return map(
-            lambda p: self.package_from_purl(p.external_references[0].locator),
+            lambda p: self.package_from_purl(self.package_manager_ref(p).locator),
             filter(self.is_debian_pkg, self._document.packages),
         )
 
     @classmethod
-    def from_file(cls, filename: Path):
+    def from_file(cls, filename: Path) -> "SpdxPackageResolver":
```
Urist-McGit, line 63: This should be in a17b55a007d7a7cced0c38f500fbb299139ad243

PR #43, opened 2025-09-25
feat: add support for compressed apt-cache lists files

Urist-McGit:
Support sources and binary package list that are compressed with bzip2, gzip, xz, zstd or lz4.

### DIFF
--- a/tests/test_generation.py
+++ b/tests/test_generation.py
@@ -5,12 +5,16 @@
 from datetime import datetime, timezone
 from importlib.metadata import version
 import json
+import os
 from pathlib import Path
+import pytest
+import subprocess
 from tempfile import TemporaryDirectory
 from urllib.parse import urlparse
 from uuid import uuid4
 
 from debsbom.apt.cache import ExtendedStates
+from debsbom.util.compression import Compression
 from debsbom.generate import Debsbom, SBOMType
 from debsbom.sbom import BOM_Standard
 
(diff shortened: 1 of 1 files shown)


src/debsbom/util/compression.py
```diff
@@ -0,0 +1,80 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+from collections import namedtuple
+from pathlib import Path
+import subprocess
+
+
+class Compression:
+    """ """
```
fmoessbauer, line 11: Empty docstring?

tests/root/apt-sources-compressed/var/lib/apt/lists/deb.debian.org_debian_dists_bookworm_main_binary-amd64_Packages.lz4
fmoessbauer, line 1: Hmm... I prefer to compress this during the test and decompress it later on. Given the xz-utils debacle, I prefer to not have binaries checked in.

tests/test_generation.py
```diff
@@ -208,6 +208,39 @@ def test_apt_source_pkg():
                     } in pkg["checksums"]
 
 
+def test_compressed_apt_source_pkg():
+    url = urlparse("http://example.org")
+    uuid = uuid4()
+    timestamp = datetime(1970, 1, 1, tzinfo=timezone.utc)
```
fmoessbauer, line 214: This becomes a pattern. Can't we either provide a factory for `Debsbom()` as a fixture, or just create the bom without manually creating the url, uuid, timestamp, etc...?

tests/test_generation.py
```diff
@@ -208,6 +208,39 @@ def test_apt_source_pkg():
...
+        sbom_types=[SBOMType.SPDX, SBOMType.CycloneDX],
+        root="tests/root/apt-sources-compressed",
+        spdx_namespace=url,
+        cdx_serialnumber=uuid,
+        timestamp=timestamp,
+    )
+
+    with TemporaryDirectory() as outdir:
+        outdir = Path(outdir)
+        dbom.generate(str(outdir / "sbom"), validate=True)
```
fmoessbauer, line 227: Nit: If we split the `.generate` and write function, we probably could write into a stream in our tests and avoid the file reading. This would also allow us to more easily emit the sbom on stdout.

src/debsbom/apt/cache.py
```diff
@@ -109,24 +111,29 @@ def _parse_sources(
         cls, sources_file: str, srcpkg_filter: Callable[[str], bool] | None = None
     ) -> Iterable["SourcePackage"]:
         try:
-            with open(sources_file) as f:
-                sources_raw = Sources.iter_paragraphs(f, use_apt_pkg=HAS_PYTHON_APT)
-                for s in Repository._make_srcpkgs(sources_raw, srcpkg_filter):
-                    yield s
-        except FileNotFoundError:
+            logger.debug(f"Parsing apt cache source packages: {sources_file}")
+            content = read_maybe_compressed_file(Path(sources_file))
```
fmoessbauer, line 115: When exactly does the compression happen? Is the file already downloaded compressed or is it compressed by apt? Does the `InRelease` file provide a compressed entry? Then we could just look for that instead of globbing / reconstructing possible patterns (possibly by even checking the checksum).

src/debsbom/util/compression.py
Urist-McGit: As copied from the initial implementation. I will remove it

tests/test_generation.py
Urist-McGit: Thats true, I will unify the tests in another MR. There is also the fact that we do not always use the pytest provided temporary directories.

Urist-McGit:
Reworked the tests to compress on the fly and fixed the bad docstring.

src/debsbom/util/compression.py
```diff
@@ -0,0 +1,79 @@
...
+        return [
+            Compression.BZIP2,
+            Compression.GZIP,
+            Compression.XZ,
+            Compression.ZSTD,
+            Compression.LZ4,
+        ]
+
+
+def read_maybe_compressed_file(path: Path) -> bytes:
```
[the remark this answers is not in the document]
fmoessbauer: Once the buffer is emptied, we need to call `.poll()` to get the return code. This probably needs to be wrapped in a class to convey the state. But IMHO this is really important to not write into some internal or OS buffers here, as we are talking about potentially many dozens of MB of uncompressed data.

PR #46, opened 2025-09-26
docs(ci): publish documentation to GitHub pages

fmoessbauer:


fmoessbauer:
> So this pushes the pages for every commit on the main branch. I am wondering if that is what we want.

Indeed. With github pages it is not trivial to version the documentation. I have looked into projects like https://real-yfprojects.github.io/sphinx-polyversion/2.0.0 but these are also not easy to setup.

For now, I propose to just publish what we have on main. Later on, we still can switch to readthedocs or build the versioning by our selves.

PR #47, opened 2025-09-26
Minor test cleanups

Urist-McGit:
Move some repeated code into functions and use tempdirs provided by pytest


Urist-McGit:
Added the fixture and it now takes a `Path`

fmoessbauer (approved):

PR #48, opened 2025-09-26
Further improvements of the documentation

fmoessbauer:


Urist-McGit:
Fair enough, we would have to maintain it in two places, which is problematic too. Since we have the documentation in github pages now it is still easily available

PR #49, opened 2025-09-26
docs(ci): download docs before publishing

fmoessbauer:
The docs are built in a separate job, hence we first need to download them before re-uploading as pages artifact.


fmoessbauer:
I tested the pages upload by temporarily adding a different branch filter.

PR #52, opened 2025-09-29
test: turn Python warnings into errors during test

fmoessbauer:


Urist-McGit:
A we do not want our pipelines to be red, I will merge this once the underlying problems that cause these warnings are fixed. See #53.

PR #65, opened 2025-10-08
Fix/improve dependency handling

tamilari:
- Show download, source-merge and repack commands in help message even
  if dependencies are not installed, so the message is consistent across
  all systems
- Display a missing dependency error message when an unavailable command
  is used
- Enforce requirement for one of the four commands to be specified, so
  only calling debsbom will display an usage error message


fmoessbauer:
Hi @tamilari,

please rebase this PR against the latest main branch, as we don't do merge commits in this project. Once done, please just force-push.

Urist-McGit:
CI is not working right now. Once #68 is merged, please rebase and I should be able to approve you.

src/debsbom/cli.py
```diff
@@ -327,7 +327,7 @@ def run(cls, args):
                 logger.warning(f"dsc file not found: {pkg.name}@{pkg.version}")
 
     @classmethod
-    def setup_parser(cls, parser):
+    def setup_parser(cls, parser, compression_options):
```
fmoessbauer, line 330: I would rather suggest to refactor this similar to the `SbomInput` by using a class (mixin) that adds and evaluates the compression argument.

PR #86, opened 2025-10-14
feat(cli): add json output option for download cmd

tamilari:
Introduce the '--json' option to the download command. When
enabled, this option provides machine-readable output in JSON format,
detailing information about each package being downloaded.

The JSON output for each file will include:
- Download status ("ok", "checksum_mismatch" or "not_found")
- Package name and version
- File name

This machine-readable format is beneficial for scripting and automation.

Note: This option is mutually exclusive with `--progress` to avoid
  interference

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 **/__pycache__
-*.json
+/*.json
 *.egg-info
 .coverage*
 docs/build/
--- a/docs/source/commands/download.rst
+++ b/docs/source/commands/download.rst
@@ -8,3 +8,35 @@
     :func: setup_parser
     :prog: debsbom
     :path: download
+
+JSON Output Schema
+------------------
+
+When the application is run with JSON output enabled (via the ``--json`` flag),
+status messages are emitted as single-line JSON objects to standard output.
+Each line represents a distinct package download operation.
+
+The schema for these JSON objects is as follows:
+
+.. literalinclude:: ../../../src/debsbom/schema/schema-download.json
+   :language: json
+
+Fields
+~~~~~~
+
+*   **status**:
+    The status of the download operation. This field will contain one of the
+    following predefined values from the :py:class:`DownloadStatus` enum:
+
+    *   ``"ok"``: The file was either successfully downloaded or found in the cache, and the checksum was verified.
+    *   ``"checksum_mismatch"``: The downloaded file's checksum did not match the expected value.
+    *   ``"not_found"``: The requested file or package could not be located.
+
+*   **package**:
+    An object identifying the package, including the name and the version as a string.
+
+*   **filename**:
+    The name of the processed file or an empty string if the package is unavailable.
+
+*   **path**:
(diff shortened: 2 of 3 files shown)


src/debsbom/commands/download.py
```diff
@@ -60,6 +60,8 @@ def _filter_pkg(p: package.Package, sources: bool, binaries: bool) -> bool:
     def run(cls, args):
         outdir = Path(args.outdir)
         outdir.mkdir(exist_ok=True)
+        if args.emit_package_status and args.progress:
```
fmoessbauer: We can already check this at parsing level using a `mutually_exclusive_group`: https://docs.python.org/3/library/argparse.html#mutual-exclusion

fmoessbauer (changes requested):

src/debsbom/commands/download.py
fmoessbauer: Good point. We probably want to make `--json` a top-level argument, as having a machine readable output might be useful for other commands as well.

src/debsbom/cli.py
```diff
@@ -87,9 +93,10 @@ def main():
             ExportCmd.run(args)
     except Exception as e:
         logger.error(e)
-        print(f"debsbom: error: {e}", file=sys.stderr)
-        if args.verbose >= 1:
-            print(traceback.format_exc(), file=sys.stderr, end="")
+        if not args.json:
```
fmoessbauer: Why? The error is anyways written to stderr.

src/debsbom/download/download.py
```diff
@@ -22,20 +27,35 @@
...
+        json: bool = False,
+        fd: io.TextIOBase = sys.stdout,
     ):
         outdir = Path(outdir)
         self.sources_dir = outdir / "sources"
         self.binaries_dir = outdir / "binaries"
         self.to_download: list[tuple[package.Package, RemoteFile]] = []
         self.rs = session
+        self.json = json
+        self.fd = fd
```
fmoessbauer: Probably not needed, as stated above.

src/debsbom/commands/download.py
```diff
@@ -85,15 +87,24 @@ def run(cls, args):
...
-        for p in dl_files:
-            logger.debug(f"downloaded {p}")
+        if not args.json:
+            nfiles, nbytes, cfiles, cbytes = downloader.stat()
+            print(
+                f"downloading {nfiles} files, {DownloadCmd.human_readable_bytes(nbytes)} "
+                f"(cached: {cfiles}, {DownloadCmd.human_readable_bytes(cbytes)})"
+            )
+
+        dl_results.extend(downloader.download(progress_cb=progress_cb if args.progress else None))
```
fmoessbauer: This already iterates (executes) the download till the end (which we need to do, but by slightly changing the logic we can already print the MR output during operation)

```python
dl_results = downloader.download(progress_cb=progress_cb if args.progress else None)

for r in dl_results:
    if (args.json):
        print(r.json())
    logger.debug(f"downloaded {r.package}")
```

PR #91, opened 2025-10-16
refactor(dpkg): add helper to separate src and bin package

fmoessbauer:
We currently have many locations in our codebase where we explicitly check for the concrete instance of package, just to see if it is a source or binary package. To avoid this antipattern, we add helpers in the package class to perform this check. We further often only want to iterate the binary packages or the source packages. For that, we add free functions to perform the filtering.


Urist-McGit:
Good changes! I never noticed how often we used this exact pattern.

PR #93, opened 2025-10-16
fix: incorrect dependency error in specific case

tamilari:
Previously, the new, clear error message for missing dependencies was only displayed if 'requests' was absent. When 'requests' was installed but 'zstandard' was missing, users would still receive a confusing error instead of the intended specific message.

This fix ensures that modules requiring 'requests' and 'zstandard' are only imported when both packages are present. This guarantees the informative missing dependency message is always shown.

### DIFF
--- a/src/debsbom/commands/download.py
+++ b/src/debsbom/commands/download.py
@@ -11,6 +11,9 @@ from ..dpkg import package
 from ..util.progress import progress_cb
 
 try:
+    # Attempt to import unused zstandard dependency to check their availability.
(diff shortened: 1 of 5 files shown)


src/debsbom/commands/download.py
```diff
@@ -12,6 +12,7 @@
 
 try:
     import requests
+    from zstandard import ZstdCompressor, ZstdDecompressor
```
fmoessbauer: Do we really need the import here? Is there any direct dependency to zstandard in this file?

PR #94, opened 2025-10-16
Add support for hierarchical merging of multiple SBOMs

Urist-McGit:
Add a new command that hierarchically merges any number of SBOMs. The merge is intended to be used for combining multiple SBOMs that describe a Debian-based Linux distribution. A common such use-case would be combining SBOMs for a rootfs and a initrd.

The merging itself takes the root components/packages of the SBOMs and moves them to the first dependency level. Any  dependencies below that are left untouched. That would mean something like this:

```
doc1-root
|- binary-dep1
|  |- source-dep1
|- binary-dep2

doc2-root
|- binary-dep3
|  |- source-dep3
|- binary-dep4
```

Would turn into this:

```
merged-doc-root
|- doc1-root
|  |- binary-dep1
|  |  |- source-dep1
|  |- binary-dep2
|- doc2-root
|  |- binary-dep3
|  |  |- source-dep3
|  |- binary-dep4
```

Duplicate components/packages are identified by their PURL. If such a component/package is found then their SBOM reference ID is replaced in the merged document. The identifiers are taken from the first occurrence in the passed SBOMs. The component/package will also be deduplicated and correctly referenced in all merged SBOMs.

### DIFF
--- a/.github/actions/pytest/action.yml
+++ b/.github/actions/pytest/action.yml
@@ -35,6 +35,14 @@ runs:
     run: |
       coverage run $(which debsbom) -v generate -t spdx -t cdx -o sbom --validate
 
+  - name: smoke test merge
+    shell: bash
+    run: |
+      export COVERAGE_FILE=".coverage.merge.spdx.${{ inputs.artifact-identifier }}"
+      coverage run $(which debsbom) -v merge --validate tests/data/merge-full.spdx.json tests/data/merge-minimal.spdx.json
+      export COVERAGE_FILE=".coverage.merge.cdx.${{ inputs.artifact-identifier }}"
+      coverage run $(which debsbom) -v merge --validate tests/data/merge-full.cdx.json tests/data/merge-minimal.cdx.json
+
   - name: test repack flow
     shell: bash
     run: |


src/debsbom/commands/input.py
```diff
@@ -67,3 +70,81 @@ class PkgStreamInput:
...
+    """
+
+    @classmethod
+    def parser_add_generate_input_args(cls, parser, default_out):
+        parser.add_argument(
+            "-t",
+            "--sbom-type",
+            choices=["cdx", "spdx"],
+            action="append",
+            help="SBOM type to generate, can be passed multiple times (default: all)",
```
fmoessbauer: As we cannot convert between cdx and spdx, passing the type on is only needed when reading the boms from stdin (which we probably anways don't support now as the importer only reads a single json object).

src/debsbom/merge/cdx.py
```diff
@@ -0,0 +1,129 @@
...
+            for component_hash in component.hashes:
+                if component_hash not in component.hashes:
+                    component.hashes.append(component_hash)
+
+    def _merge_dependency(self, dependency: Dependency, other: Dependency):
+        for dep in other.dependencies:
+            if dep not in dependency.dependencies:
+                dependency.dependencies.add(dep)
+
+    def merge_sboms(
```
fmoessbauer: Why not simply `merge`?

src/debsbom/merge/cdx.py
```diff
@@ -0,0 +1,129 @@
...
+        self, sboms: Iterable[Bom], progress_cb: Callable[[int, int, str], None] | None = None
+    ) -> Bom:
+        logger.info("Merging SBOMs")
```
fmoessbauer: Is this valuable information? If not, better remove the log message.

src/debsbom/merge/cdx.py
```diff
@@ -0,0 +1,129 @@
...
+        root_bom_refs = []
+        ref_map = {}
+
+        num_steps = 0
+        cur_step = 0
+        if progress_cb:
+            for sbom in sboms:
+                num_steps += len(sbom.components) + len(sbom.dependencies)
+
+        for sbom in sboms:
```
fmoessbauer: We pass sboms as an `Iterable`, so we can't iterate it twice. Either pass it as list, or just iterate once.

tests/test_merge.py
```diff
@@ -0,0 +1,101 @@
...
+from debsbom.merge.spdx import SpdxSbomMerger
+
+
+def test_spdx_merge():
+    distro_name = "spdx-merge-package-merge"
+    merger = SpdxSbomMerger(distro_name=distro_name)
+    docs = []
+    for sbom in ["tests/data/merge-full.spdx.json", "tests/data/merge-minimal.spdx.json"]:
+        docs.append(SpdxBomReader.read_file(Path(sbom)))
+    bom = merger.merge_sboms(iter(docs))
```
fmoessbauer: If we had a progress reporting function setup, this would already have failed due to the double iteration.

fmoessbauer (commented):
In general, this already looks pretty good. While we work on the last remaining pieces, feel free to move all commits except for the last two into a separate MR. Then we can already merge them while we polish here.

PR #101, opened 2025-10-21
Correct automodule paths for commands

Urist-McGit:
The automodule paths were incorrect after the recent refactoring of commands. As no docstrings were provided in any of the command modules no documentation was actually missed.


fmoessbauer (approved):
Good catch!

PR #103, opened 2025-10-22
chore(resolver): on imprecise lookup return most-recent files

fmoessbauer:
When doing a lookup of a source package, whereby the source package has multiple name / version identical entries on the snapshot mirror (e.g. because it was re-signed over time without changing the content), we now only return the most-recent version. By that, we avoid the name collisions in the download directory. We further can't do any better, as we do not know which version was used when generating the SBOM.

The case where we have precise data (e.g. a checksum) is not changed. There, we still iterate all artifacts and return the one with matching checksum.

### DIFF
--- a/src/debsbom/download/resolver.py
+++ b/src/debsbom/download/resolver.py
@@ -101,7 +101,29 @@ class PersistentResolverCache(PackageResolverCache):
 
 class UpstreamResolver:
     """
-    Helper to lookup packages on an upstream snapshot server.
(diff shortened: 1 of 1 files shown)


Urist-McGit:
I think we are getting to a point where the whole lookup needs some documentation on how it is done and where caveats are. We always handle packages with checksums correctly, but packages without checksums can never be identified with 100% accuracy. We take a good guess there, so that should be documented

PR #105, opened 2025-10-22
fix(cli): print json output only with --json

tamilari:
Fixes: 16ce863 ("feat(cli): add --json option for download cmd")


fmoessbauer (approved):
Thanks

PR #106, opened 2025-10-22
fix(reference): correcly resolve dependencies of arch-all packages

fmoessbauer:
Architecture all packages can depend on architecture native packages. This was missed when adding the multi-arch support, as only the direction arch=<...> -> arch=all was mapped but not the other way round. As we might have multiple architecture specific dependencies with the same name and version at a time, we need to explicitly resolve to the native arch. We now read this from the rootfs if not explicitly specified by the caller.


src/debsbom/commands/generate.py
```diff
@@ -78,3 +79,9 @@ def setup_parser(cls, parser):
             help="create SBOM from a package list passed via stdin",
             action="store_true",
         )
+        parser.add_argument(
+            "--distro-arch",
+            type=str,
+            help="native dpkg architecture of the distro (default: %(default)s)",
+            default="amd64",
```
Urist-McGit: We use a default here, but the parsing of the dpkg native arch only happens if nothing is passed. So the user can never pass anything here to make it auto-detect. I think the default should be nothing, we auto-detect, and if that fails the user has to explicitly give an architecture

PR #108, opened 2025-10-23
Fix merging of checksums for merge command

Urist-McGit:
The merging of checksums for components/packages was not done correctly for both formats. Additionally when the checksums mismatch we raise an error now. Add tests for these cases too.

### DIFF
--- a/tests/data/checksum-merge-bad.cdx.json
+++ b/tests/data/checksum-merge-bad.cdx.json
@@ -1,0 +1,67 @@
+{
+    "components": [
+        {
+            "bom-ref": "pkg:deb/debian/example-pkg@1.0.0?arch=amd64",
+            "group": "debian",
+            "name": "example-pkg",
+            "properties": [
+                {
+                    "name": "section"
+                }
+            ],
+            "purl": "pkg:deb/debian/example-pkg@1.0.0?arch=amd64",
+            "type": "library",
+            "version": "1.0.0",
+            "hashes": [
+                {
+                    "alg": "MD5",
+                    "content": "37f5724421f3cfedfa0ceaf9b48473dd"
+                },
+                {
+                    "alg": "SHA-256",
+                    "content": "10b684e6d5a68698e9370585285f3bf858aecea6f25f70453bc9b204daa7d6f1"
+                }
+            ]
+        }
+    ],
+    "dependencies": [
+        {
+            "dependsOn": [
+                "pkg:deb/debian/example-pkg@1.0.0?arch=amd64"
+            ],
+            "ref": "CDXRef-checksum-merge-md5"
+        },
+        {
+            "ref": "pkg:deb/debian/example-pkg@1.0.0?arch=amd64"
+        }
+    ],
+    "metadata": {
+        "component": {
+            "bom-ref": "CDXRef-checksum-merge-md5",
+            "name": "md5",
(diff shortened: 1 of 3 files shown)


src/debsbom/merge/cdx.py
```diff
@@ -8,6 +8,7 @@
 from cyclonedx.model.dependency import Dependency
 import itertools
```
fmoessbauer: Please add a fixes tag

src/debsbom/merge/cdx.py
```diff
@@ -28,12 +29,11 @@ def _merge_component(self, component: Component, other: Component):
...
-            if component.hashes is None:
-                component.hashes = []
-            for component_hash in component.hashes:
-                if component_hash not in component.hashes:
-                    component.hashes.append(component_hash)
+        if component.hashes is None:
+            component.hashes = SortedSet([])
+        for component_hash in other.hashes:
+            if component_hash not in component.hashes:
+                component.hashes.add(component_hash)
```
fmoessbauer: IMHO the merging is more complicated: If we have a checksum in both components, these must match. Otherwise the the components cannot be merged (either we raise an exception, or just file a warning and take the first component). This needs to be checked for each checksum.

fmoessbauer: As we were told in [1], this assumption does not always hold (it should hold on semantic level, but not on bit-level). I'm wondering why this should be complicated to implement. Just iterate all hashes of the self package and compare with matching entries from the other package before merging the hashes.

[1] https://lists.debian.org/debian-snapshot/2025/10/msg00002.html

src/debsbom/merge/cdx.py
```diff
@@ -23,20 +23,33 @@ def _merge_component(self, component: Component, other: Component):
         # merge all fields that we use in our SBOMs, all other fields
         # are either part of the PURL or the PURL itself, and these
         # must match before merging
+
+        if component.hashes is None:
+            component.hashes = SortedSet([])
+        if other.hashes:
```
fmoessbauer: Avoid one level of cyclomatic complexity by using `for ... in other.hashes or []`. But I'm also wondering why we iterate the sorted set (why sorted?) and not simply the dict. Then we could just check against the other set with `other.hash.get(algo)`. For me, this code is really hard to read.

src/debsbom/merge/spdx.py
```diff
@@ -44,17 +45,31 @@ def _hash_relationship(cls, rel: Relationship) -> int:
...
-        if package.supplier == SpdxNoAssertion():
-            package.supplier = other.supplier
 
         if package.checksums is None:
             package.checksums = []
         if other.checksums:
-            for checksum in other.checksums:
-                if checksum not in package.checksums:
-                    package.checksums.append(checksum)
+            for other_chksum in other.checksums:
```
fmoessbauer: Same thing here. `for other_chksum in other.checksums or []` to avoid one level of indentation.

src/debsbom/merge/cdx.py
Urist-McGit: CDX expects a `SortedSet` in the field, passing anything else causes an error. I could create a list first and convert it later, but this seemed the easier way.

What dict are you refererring to?

fmoessbauer (approved):

PR #109, opened 2025-10-23
chore(dpkg): add md5sum of source package dsc files as well

fmoessbauer:
Apt provides this data, so add it to the SBOM as well.


Urist-McGit:
`feat(test): check all pkg checksums we get from the apt-cache` should be `chore(..)` not `feat(..)`.

fmoessbauer:
> `chore`

Conventional changelog on tests... I'm still guessing. Fixed it as proposed.

Issue #100, opened 2025-10-21
reproducible result for `source-merge`


gernot-h:
In some workflows, reproducible results (i.e. with identical hash) for source archives are very helpful to verify known sources and avoid duplicated uploads. 

Following https://reproducible-builds.org/docs/archives/, I therefore decided to use the following commandline in an internal predecessor tool for Debian source downloads:

```python
subprocess.check_call([
                "tar", "cjf", filename, "--force-local",
                "--format=gnu", "--sort=name",
                "--owner=0", "--group=0",
                "--numeric-owner", "--mtime=2020-01-01 00:00:00Z",
                unpack_dir])
```

The same commandline  was also [adopted by the Continuous Clearing Tool from colleagues](https://github.com/siemens/continuous-clearing/blob/v8.2.1/src/LCT.SW360PackageCreator/DebianPatcher.cs#L27). 

In general, it would be good if debsbom would create reproducible source archives and if it would allow to create the same results than our current tools, this would ease our migration to debsbom. ;-)

PR #111, opened 2025-10-24
Fm/tar reproducible

tamilari:


Urist-McGit:
For compatibility reasons (#100) it might also be useful to have the timestamp as option for the CLI.

src/debsbom/util/compression.py
```diff
@@ -12,12 +12,12 @@ class Compression:
...
-    GZIP  = Format("gzip",  ["-q"], ["-q", "-d", "-c"], ".gz")
-    XZ    = Format("xz",    ["-q"], ["-q", "-d", "-c"], ".xz")
-    ZSTD  = Format("zstd",  ["-q"], ["-q", "-d", "-c"], ".zst")
-    LZ4  = Format("lz4",    ["-q"], ["-q", "-d", "-c"], ".lz4")
+    NONE  = Format("cat",   [],           [],                 "")
+    BZIP2 = Format("bzip2", ["-q"],       ["-q", "-d", "-c"], ".bz2")
+    GZIP  = Format("gzip",  ["-q", "-n"], ["-q", "-d", "-c"], ".gz")
+    XZ    = Format("xz",    ["-q"],       ["-q", "-d", "-c"], ".xz")
+    ZSTD  = Format("zstd",  ["-q"],       ["-q", "-d", "-c"], ".zst")
+    LZ4   = Format("lz4",   ["-q"],       ["-q", "-d", "-c"], ".lz4")
```
Urist-McGit: You messed up the formatting here I think? Only line 17 should change

src/debsbom/repack/merger.py
```diff
@@ -78,15 +80,36 @@ def locate_artifact(cls, p: package.Package, basedir: Path) -> Path | None:
...
+            if not d.is_dir():
+                continue
+            cand = d / "debian" / "changelog"
+            if cand.is_file():
+                changelog_path = cand
+                break
+        if not changelog_path:
+            return None
+        # Open and parse the changelog
+        with changelog_path.open() as changelog_file:
```
fmoessbauer: Nit: in the rest of the codebase, we use the free `open` function instead of the one from pathlib. Maybe switch here as well.

src/debsbom/commands/repack.py
```diff
@@ -96,3 +96,7 @@ def setup_parser(cls, parser):
             help="validate generated SBOM (only for SPDX)",
             action="store_true",
         )
+        parser.add_argument(
+            "--mtime",
+            help="set mtime for creating tar archives",
```
Urist-McGit: We should mention that this argument is passed to directly to `tar`, otherwise it might not be clear what time formats are allowed here.

src/debsbom/repack/merger.py
```diff
@@ -41,6 +41,7 @@ def __init__(
     ):
         self.dldir = dldir
         self.outdir = outdir or dldir
+        self.outdir.mkdir(exist_ok=True, parents=True)
```
fmoessbauer: We usually use `parents=False`, as recursively creating directories is dangerous (or at least easily leads to unexpected results).

src/debsbom/repack/merger.py
```diff
@@ -106,7 +107,9 @@ def merge(
         dsc = self.locate_artifact(p, self.dldir)
         if not dsc:
             raise DscFileNotFoundError(p.dscfile())
-        merged = dsc.with_suffix(suffix)
+        dir = self.outdir / dsc.parent.name
+        dir.mkdir(exist_ok=True)
+        merged = dir / dsc.with_suffix(suffix).name
```
fmoessbauer: Why change to `.name` here?

tests/conftest.py
```diff
@@ -22,11 +22,14 @@
 beartype_package("debsbom", conf=bt_conf)
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
 def http_session():
     rs = requests.Session()
     rs.headers.update({"User-Agent": f"debsbom/{version('debsbom')}+test"})
-    return rs
+    try:
```
fmoessbauer: Can't we simply use a context manager here? By that, we can avoid the try / finally pattern.

src/debsbom/repack/merger.py
```diff
@@ -78,15 +80,36 @@ def locate_artifact(cls, p: package.Package, basedir: Path) -> Path | None:
                 return cand
         return None
 
-    def merge(self, p: package.SourcePackage, apply_patches: bool = False) -> Path:
+    @staticmethod
+    def extract_timestamp(path: Path) -> str | None:
```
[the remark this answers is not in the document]
Urist-McGit: If we use a structured approach here we should just use `datetime`.

src/debsbom/commands/repack.py
Urist-McGit: See in the discussion below, it would be better if we accept only ISO 8601 strings here. We can directly parse this into a `datetime` and fail early if the passed time is bad. This is already done like this in the `generate` command with the `--timestamp` option

src/debsbom/util/compression.py
Urist-McGit: Ah, true. I am not a fan of these formatted blocks for that reason, but it can stay like that for now.

src/debsbom/repack/merger.py
tamilari: To append the file name (and not the full absolute/relative path) to the path before the '/'. But is there a better way?

fmoessbauer: Ah ok, then it is fine.

tests/test_source_merger.py
```diff
@@ -57,11 +61,62 @@ def some_packages(dldir, http_session):
...
     sam = SourceArchiveMerger(dldir / "sources", outdir, compress=Compression.from_tool(compress))
 
     for p in some_packages:
-        assert p.name in sam.merge(p, apply_patches=True).name
+        if compress != "gzip":
+            expected_timestamp_str = "Wed, 01 Oct 2025 12:34:56 +0100"
+        else:
+            expected_timestamp_str = EXPECTED_CHANGELOG_TIMESTAMPS[(p.name, p.version)]
+        try:
+            dt_object = datetime.strptime(expected_timestamp_str, "%a, %d %b %Y %H:%M:%S %z")
```
fmoessbauer: Isn't that the ISO 8601 format? For that, you can use `datetime.fromisoformat()` to parse it.

tests/test_source_merger.py
```diff
@@ -57,11 +61,62 @@ def some_packages(dldir, http_session):
...
+            )
+        result = sam.merge(
+            p,
+            apply_patches=True,
+            mtime=None if compress != "bzip2" else dt_object,
+        )
+        assert p.name in result.name
+
+        if compress != "bzip2" and compress != "gzip":
+            continue  # skip timestamp checks for other compression types to reduce test time
```
fmoessbauer: Please don't skip the test. If performance is an issue, rather use smaller packages.

tests/test_source_merger.py
```diff
@@ -57,11 +61,62 @@ def some_packages(dldir, http_session):
...
 @pytest.mark.parametrize("compress", [None, "bzip2", "gzip", "xz", "zstd", "lz4"])
 @pytest.mark.online
 def test_merger(tmpdir, some_packages, dldir, compress):
     outdir = Path(tmpdir / "merged")
     sam = SourceArchiveMerger(dldir / "sources", outdir, compress=Compression.from_tool(compress))
 
     for p in some_packages:
-        assert p.name in sam.merge(p, apply_patches=True).name
+        if compress != "gzip":
+            expected_timestamp_str = "Wed, 01 Oct 2025 12:34:56 +0100"
```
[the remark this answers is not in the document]
tamilari: Initially, I considered testing one compression type with a fixed mtime and another with the latest timestamp from the changelog. However, we now do both with every compression type. As you said, if the performance is too poor, we'll use smaller packages.

PR #118, opened 2025-10-29
Actually remove packages that are not properly installed from generated SBOMs

Urist-McGit:
Packages that are not properly installed but have some residual config still show up in the dpkg status file. We do not want to place these packages into the SBOM. As it turns out we never properly removed them but instead only disabled dependency  resolution. Actually remove them now.


Urist-McGit:
The new changes actually filter out all binary and source packages that would come from non-installed packages. The changes go deep in the parsing of packages instead of in the generation step. I also expanded the test to check for source packages that might be pulled in by non-installed packages.

fmoessbauer (approved):
Thanks for fixing this.

PR #120, opened 2025-11-03
docs: document design decisions

### LINKED ISSUE #113: Document SBOM design decisions
Many design decisions are currently not documented, or only implicitly somewhere in the code. Many of these could be considered implementation details, but they are very relevant for consumers, and so these should be documented so they can be relied upon.

What should be documented:

- What ends up in the SBOM
  The mapping of fields in the Debian package representation to both SBOM formats
- source - binary - built-using
  How dependencies between these packages are mapped to SBOMs formats
- component type / primary package purpose
  Why we use the component type/package purpose `library`
- checksums: source -> dsc; binary -> .dep
  What the checksum of a package/component describes
- what makes a component/package unique
  Uniqueness of binary and source packages is surprisingly complicated. How do we identify and express uniqueness for components/packages
- package/component identification
  Related to the previous point, how doe we identify which package is meant in an input SBOM
- distro-arch
  Resolving dependencies requires knowledge of the architecture of the described operating system
- minimal package/component contents
  Many of the fields in SBOMs are optional and will only be filled if the data is available. There are some fields which are always filled though
- checksum matching
  How we handle checksums, both for verification and what we do in their absence
- universal ingress apt-cache handling
  When passing a package list there is some special behavior when it comes to apt-cache parsing and package/component enrichment
-  we dont guess philosophy
- what data do we use (list all the files)
  We parse a number of files from dpkg and apt
- what is offline what is online
  Some commands are usable completely offline, others require internet access

Urist-McGit:
In the development of debsbom a number of design decisions was made. Document these so they can be referenced, and users can understand and know which parts of the behavior of the tool can be relied upon, and which is just incidental.

Depends on #119 , since it references some of the checksum matching code.

Closes #113

### DIFF
--- a/docs/source/intro.rst
+++ b/docs/source/intro.rst
@@ -36,8 +36,8 @@ Non-Goals
 - Real-time vulnerability database integration
 - Signing and attestation of generated artifacts
 
-Package Relations
------------------
+Package Relationships
+---------------------
 
 A Debian distribution is composed of source packages and binary packages.
 Binary packages are installed into the root filesystem, while the source packages are the originals from which those binaries are built.
@@ -45,7 +45,7 @@ Binary packages are installed into the root filesystem, while the source package
 Some binary packages are installed explicitly by the user; others appear automatically as dependencies of the explicitly‑installed packages.
 The SBOM mirrors this relationship, using the ``distro-package`` entry as the single central node for traversing the package graph.
 
-Relations hierarchy::
+Relationship hierarchy::
 
     distro-package
     ├─ binary-package-foo


docs/source/design-decisions.rst
```diff
@@ -0,0 +1,195 @@
...
+For SPDX SBOMs it is possible to differentiate them: we use ``primaryPackagePurpose = LIBRARY`` for binary packages and ``primaryPackagePurpose = SOURCE`` for source packages.
+
+Checksums
+---------
+
+``debsbom`` includes checksums in components/packages. For binary packages these refer to the checksum of the ``.deb`` file.
+
+The situation is more complicated for source packages, as these consist of multiple files and there is no checksum given for the collection of them. ``debsbom`` uses the checksum of the ``.dsc`` file in each source package, as this file describes the whole source package and its contents, together with checksums for all other included files. This makes it possible to verify that every file in a source package is as expected.
+
+When there are multiple checksums of different algorithms available ``debsbom`` checks them in a specific internal order. It tries to match from highest to lowest priority, and a single match is considered enough. The order is based on cryptographic strengths of the algorithm. The order is from highest to lowest ``SHA256``, ``SHA1``, ``MD5``.
```
fmoessbauer: A single match is enough is too imprecise. The first algorithm that is supported is checked. The result of this check is final.

docs/source/design-decisions.rst
```diff
@@ -0,0 +1,195 @@
...
+
+The goal of this page is to explicitly document and give reasoning for design decisions that were made in this project.
+
+"We don't guess" Philosophy
+---------------------------
+
+The philosophy for ``debsbom`` when accruing information from a distribution is simple: "We don't guess". That means that ``debsbom`` always prefers providing less information to providing unverified information. In other words: if anything in the SBOM is wrong, it is a bug.
+
+Offline SBOM Generation
+--------------------------
```
fmoessbauer: This line is off

PR #122, opened 2025-11-05
chore(cache): gracefully parse deb822 cache data

fmoessbauer:
We observed cases in the field where a third party debian repo provider (in this case Artifactory) generates incorrect deb822 data. This error remained unnoticed, as apt itself does not extract the incorrect fields.

This patch makes the parser gracefully degradate by skipping incorrect entries but continuing anyways.

### DIFF
--- a/tests/root/illformed-sources/source_Sources
+++ b/tests/root/illformed-sources/source_Sources
@@ -1,0 +1,94 @@
+Package: argon2
+Binary: libargon2-dev, libargon2-0, libargon2-1, argon2, libargon2-1-udeb
+Version: 0~20171227-0.3+deb12u1
+Priority: optional
+Section: misc
+Maintainer: Luca Bruno <lucab@debian.org>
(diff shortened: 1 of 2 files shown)


src/debsbom/apt/cache.py
```diff
@@ -107,7 +107,10 @@ def _make_srcpkgs(
             else sources
         )
         for source in _sources:
-            yield SourcePackage.from_deb822(Dsc(source))
+            try:
+                yield SourcePackage.from_deb822(Dsc(source))
+            except KeyError:
```
Urist-McGit: Is it enough to rely only on `KeyError`? If the format is wrong then I would expect a multitude of exceptions depending on what exactly is malformed.

Also debug logging the encountered exception might be useful for debugging purposes.

fmoessbauer: Now we debug-log the exception as well.

PR #124, opened 2025-11-07
Fm/vcs link

tamilari:
feat: add Vcs informations to sbom

For spdx, add the Vcs information in the 'download_location' field. Format this string according to the spdx documentation. It is not possible to add the url to the web interface for browsing the repository, since there is no field for this information.

For cdx add the Vcs information and the url to the web interface to the 'external_refs' object with the type set to 'vcs'.

This Vcs information can be useful for analysing the packages (e.g. how well a package is maintained and what the risk factor is).

### DIFF
--- a/tests/test_generation.py
+++ b/tests/test_generation.py
@@ -192,6 +192,14 @@ def test_apt_source_pkg(tmpdir, sbom_generator):
                     "algorithm": "SHA256",
                     "checksumValue": "cd75da7829d819189ba6154d408666373b307e222b393223804c4c4a7156f421",
                 } in pkg["checksums"]
+                externalRefs = pkg["externalRefs"]
+                assert len(externalRefs) == 2
+                for ref in externalRefs:
+                    if ref["referenceType"] == "vcs":
+                        assert (
+                            ref["referenceLocator"]
+                            == "https://salsa.debian.org/toolchain-team/binutils.git"
+                        )
             if pkg["SPDXID"].endswith("binutils-arm-none-eabi-amd64"):
                 assert {
                     "algorithm": "MD5",
@@ -201,6 +209,16 @@ def test_apt_source_pkg(tmpdir, sbom_generator):
                     "algorithm": "SHA256",
                     "checksumValue": "c8f9da2a434366bfe5a66a8267cb3b1df028f1d95278715050c222b43e1c221c",
                 }
+    with open(outdir / "sbom.cdx.json") as file:
+        spdx_json = json.loads(file.read())
+        packages = spdx_json["components"]
+        for pkg in packages:
+            if pkg["bom-ref"].endswith("binutils@2.40-2?arch=source"):
+                externalRefs = pkg["externalReferences"]
+                assert len(externalRefs) == 2
+                for ref in externalRefs:
+                    if ref["type"] == "vcs":
+                        assert ref["url"] == "https://salsa.debian.org/toolchain-team/binutils.git"
 
 
 def test_apt_pkgs_stream(tmpdir, sbom_generator):


src/debsbom/generate/cdx.py
```diff
@@ -64,20 +64,42 @@ def cdx_package_repr(
...
                 type=cdx_model.ExternalReferenceType.WEBSITE,
                 comment="homepage",
             ),
         )
+
     if package.is_binary():
         entry.description = package.description
         entry.properties.add(cdx_model.Property(name="section", value=package.section))
+        if external_refs:
+            entry.external_references = external_refs
```
fmoessbauer: If these can be none as well, we don't need the if.

src/debsbom/generate/cdx.py
```diff
@@ -64,20 +64,42 @@ def cdx_package_repr(
...
+                    type=cdx_model.ExternalReferenceType.VCS,
+                    comment="URL of a web interface for browsing the repository",
+                ),
+            )
+        if package.vcs_type and package.vcs:
+            external_refs.append(
+                cdx_model.ExternalReference(
+                    url=cdx_model.XsUri(package.vcs),
+                    type=cdx_model.ExternalReferenceType.VCS,
+                    comment=f"Version control system of type {package.vcs_type.value}",
```
fmoessbauer: Does this comment really help? Are these external references allowed to be imprecise, or are further semantics linked to them (i.e. they are download locations for the exact package sources)?

src/debsbom/generate/spdx.py
```diff
@@ -90,6 +91,36 @@ def make_creation_info(
...
+    branch = re.search(r"\s-b\s(\S+)", vcs)
+    if branch:
+        download_location += f"@{branch.group(1)}"
+
+    # match "[subdir]"
+    subdir = re.search(r"\s\[(\S+)\]", vcs)
+    if subdir:
+        download_location += f"#{subdir.group(1)}"
+
+    return download_location
```
fmoessbauer: Are the entries in the .dsc files (and alike) usually precise, or do they just link to the upstream repo? My interpretation of https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field is that it is in general allowed to be imprecise, but it I would also like to hear the opinion of @gernot-h .

tests/test_generation.py
```diff
@@ -346,3 +347,47 @@ def test_illformed_sources():
...
+        (
+            VcsInfo(
+                VcsType.Git,
+                "git+ssh://git.myproject.org/MyProject [lib/variable.rb] -b da39",
+            ),
+            "git+ssh://git.myproject.org/MyProject@da39#lib/variable.rb",
+        ),
+    ],
+)
+def test_format_vcs(input, expected):
```
Urist-McGit: Could we also have a test that verifies the parsing works as expected? The setup should be available in the apt-sources root, from there you could simply check if it parsed correctly into a package.

src/debsbom/generate/spdx.py
tamilari: I also thought that this field wasn't the best fit for VCS information, but the description and examples provide such detailed instructions on how to format the VCS information. I couldn't find a better place for it, and I assume that we need this information in the SPDX format as well. Alternatively, maybe it could go in the source information field (https://spdx.github.io/spdx-spec/v2.3/package-information/#712-source-information-field)?

Urist-McGit: Yes, I would do something like `category = Other`  and `type = vcs`

src/debsbom/generate/cdx.py
Urist-McGit: This helps identifying which vcs type is actually used. It should mostly be git, but could be others, like Mercurial, SVN etc. That information might not be immediately apparent from the URL, so it is useful.

src/debsbom/generate/spdx.py
tamilari: As far as I know, we sadly don't have the link to the dsc file at the time of creating the SBOMs, since the `generate` command operates fully offline.

tamilari: > Thank you! While I'm no SPDX expert, the CDX part looks good to me! I just wonder whether we really need the "web interface for browsing the repository" part, but I also couldn't find anything in the CDX spec or guide which would suggest leaving it out either.

I am also unsure if it provides any additional value, since in 99% of the cases the only difference between the two links is a missing `.git` at the end. And if we let it in, maybe the `website` type would be better for CDX?

tamilari:
Okay, I've removed the "vcsbrowser" information. Thank you for your help!

PR #125, opened 2025-11-07
Refactor/minor

tamilari:
There were two minor things that I was a little confused about. But there might be a reason for them?


fmoessbauer (approved):
Makes sense. Thanks!

PR #127, opened 2025-11-14
fix(repack): correct apply-patches flag handling and exclude upstream tarb…

ssnaaz:
…alls

The --apply-patches flag was not propagated to the packer, causing dpkg-source to always use --skip-patches and omit the .pc directory in merged source archives. Additionally, the repack step included the upstream .orig.tar.* file alongside the extracted source tree.

Forward the flag correctly and ensure only the unpacked source directory is included in the merged tarball.

### DIFF
--- a/src/debsbom/commands/repack.py
+++ b/src/debsbom/commands/repack.py
@@ -41,7 +41,11 @@ class RepackCmd(SbomInput, RepackInput):
             pkg_subset = None
 
         packer = Packer.from_format(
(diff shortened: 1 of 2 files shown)


fmoessbauer:
Hi @ssnaaz , thanks for the fix. Please just format the code with black and add a fixes tag to your commit message (amend the commit and then force push). The following commit probably introduced the bug, hence the fixes tag should be:

```
Fixes: b1837f8 ("feat: add command to create a standard bom ...")
```

PR #129, opened 2025-11-19
Add plugin support for download resolving

Urist-McGit:
Add a new `--resolver` option to the `download` command. New choices for this option can be added with plugins. The plugin functionality is based on entry points.

~This is marked as Draft since I consider a full example of a plugin necessary documentation, and for technical reason this has to live in a separate repository. Until the Siemens internal review process for open source publications is done I can not publish it, so we have to wait until then. Otherwise I consider this ready, except for the missing link to the plugin examples repository.~
The example repository can be found here: https://github.com/Urist-McGit/debsbom-plugin-examples

### DIFF
--- a/src/debsbom/commands/download.py
+++ b/src/debsbom/commands/download.py
@@ -117,7 +117,8 @@ class DownloadCmd(SbomInput, PkgStreamInput):
                 DownloadCmd._check_for_dsc(pkg, files)
                 downloader.register(files, pkg)
             except ResolveError:
-                logger.warning(f"not found upstream: {pkg}")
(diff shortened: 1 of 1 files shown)


src/debsbom/commands/download.py
```diff
@@ -134,3 +141,9 @@ def setup_parser(cls, parser):
             metavar="SKIP",
             help="packages to exclude from the download, in package-list format",
         )
+        parser.add_argument(
+            "--resolver",
+            choices=RESOLVERS.keys(),
+            default=next(iter(RESOLVERS.keys())),
```
fmoessbauer: Nit: We probably better select our resolver explicitly.

PR #130, opened 2025-11-21
fix(repack): enforce mode=a=rX,u+w to avoid permission discrepancies

ssnaaz:
Archives created on different systems had varying file permissions, leading to different checksums after repacking. Setting mode=a=rX,u+w to normalize permissions to ensure reproducibility.

Fixes: 6c54581 (chore(repack): create tar with stable order and owner)

### DIFF
--- a/src/debsbom/repack/merger.py
+++ b/src/debsbom/repack/merger.py
@@ -185,6 +185,7 @@ class SourceArchiveMerger:
                 "--force-local",
                 "--format=gnu",
                 "--sort=name",
+                "--mode=a=rX,u+w",
                 "--owner=0",
                 "--group=0",
                 "--numeric-owner",


Urist-McGit:
The commit message should be `fix(repack): enforce mode=755 to avoid permission discrepancies`

ssnaaz:
@Urist-McGit Updated, thanks!

fmoessbauer (changes requested):
This makes all files executable. We instead need either `--mode=a=rX,u+w` (1) or `--mode=go+u,go-w` (2)

1. https://reproducible-builds.org/docs/archives/
2. https://www.gnu.org/software/tar/manual/html_section/Reproducibility.html

fmoessbauer:
> Thanks @fmoessbauer, I have verified both options: `--mode=a=rX,u+w` and `--mode=go+u,go-w` and both work as expected. However, I chose the first option (`a=rX,u+w`) because it best addresses our issue: on system A the permission is `rw-rw-r--`, whereas on system B it is `rw-r--r--`, ensuring reproducible file modes across systems. Please let me know if my assumption is correct and kindly review this change.

That makes sense. Thanks. Merged.

PR #132, opened 2025-11-25
feat(download): add session adapter for local files

Urist-McGit:
Add an adapter to the request session to allow access to local files. Any URLs that start with `file:///` will be picked up by this adapter and copied locally.


fmoessbauer:
May I ask you to also add a test?

PR #134, opened 2025-11-28
Refactor/optional SBOM dependencies

tamilari:
Previously, both libraries were always required, even if a user only intended to use one format, which would lead to errors if the unused dependency was missing. Now, users only need to install the specific dependency for the SBOM format they wish to use. This also means that missing dependency errors are now more precise, occurring only when a user tries to execute a command for a format whose required library is not installed.
This update may simplify packaging efforts for distributions like Debian, as `debsbom` can now be packaged with just `python3-cyclonedx-lib` even if a `spdx-tools` package is not yet available.

### DIFF
--- a/docs/source/starting.rst
+++ b/docs/source/starting.rst
@@ -14,22 +14,25 @@ Virtual Environment
 
    ``source <env_name>/bin/activate``
 
-3. Install the dependencies in the virtual environment with one of the following commands:
+3. Install debsbom and its dependencies in the virtual environment with one of the following commands:
 
-   - ``pip3 install -e .``
-        Installs the dependencies for all but the download command.
+   - Installs the debsbom pip package with its core dependencies:
 
-   - ``pip3 install -e .[download]``
-        Installs the dependencies for all commands.
+      ``pip3 install debsbom[cdx,spdx,download]``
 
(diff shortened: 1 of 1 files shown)


Urist-McGit:
Nit: 2cba7a106983abaf52698b79e84822f80f6e8ea9 is not really a fix. It worked as intended before and is just a style issue since we take the reexport instead of the direct one. So I would suggest to change it to `chore(export)` and remove the fixes tag

src/debsbom/commands/generate.py
```diff
@@ -23,9 +23,23 @@ class GenerateCmd(GenerateInput):
...
+            for stype in [SBOMType.SPDX, SBOMType.CycloneDX]:
+                try:
+                    stype.validate_dependency_availability()
+                    sbom_types.append(stype)
+                except RuntimeError:
+                    pass
+            if sbom_types == []:
+                raise RuntimeError(
+                    "Cannot generate any SBOM because no SBOM format dependencies are available. "
+                    "Install them by enabling the dependency specifiers `cdx` and/or `spdx`: "
```
Urist-McGit: It is not `specifiers` but `extras` here. I gave the wrong string on the initial suggestion, mea culpa.

PR #135, opened 2025-11-28
Test/test cdx and spdx individually

tamilari:
In my opinion, it is really valuable to test cdx and spdx individually. ~But I don't really like the idea of uninstalling a specific package, neither only testing one individually (and the other one with both dependencies installed), nor reinstalling everything. Any opinions?~

Depends on #134

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,27 @@ on:
   pull_request:
 
 jobs:
+  test-minimal-deps:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        sbom-type: ["cdx", "spdx"]
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -11,6 +11,28 @@ on:
...
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        sbom-type: ["cdx", "spdx"]
+        python-version: ["3.13"]
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
```
Urist-McGit: We only use a single python version for the test, so no need to keep it in a matrix.

PR #143, opened 2025-12-03
docs(examples): update From Package List examples to include --distro…

asrinuva:
…-arch

SBOM generation via the From Package List method mandates providing the distribution architecture through --distro-arch. This update adds the missing detail to the examples to make it clear.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -37,11 +37,11 @@ Create the SBOM from a package list. The so provided packages will still be enri
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


asrinuva:
@fmoessbauer Please have a look, thanks.

asrinuva:
> ere are two use-cases for package list ingress

Okay, thanks for the explanation. IIUC, when generating an SBOM from a package list (via stdin), SBOM by default reads the host’s rootfs (/) apt-cache to retrieve package information. Any packages specified can then be enriched with the data found there, which also facilitates automatic detection of the distro architecture.

PR #146, opened 2025-12-05
feat(generate): include snapshot date in SBOM metadata

asrinuva:
Add the snapshot date to the SBOM metadata to facilitate downstream layers identify which snapshot the packages were sourced from. This allows them to regenerate the SBOM using the same snapshot for comparison when submitting additional package clearance.

### DIFF
--- a/src/debsbom/generate/generate.py
+++ b/src/debsbom/generate/generate.py
@@ -6,6 +6,7 @@ from collections.abc import Callable, Iterable
 from datetime import datetime
 from io import TextIOWrapper
 import itertools
+import re
 import sys
 import logging
 from pathlib import Path
@@ -21,6 +22,7 @@ from ..dpkg.package import (
 from ..bomwriter import BomWriter
 from ..sbom import SBOMType, BOM_Standard
 
+SNAPSHOT_RE = re.compile(r'(\d{8}T\d{6}Z)')
 
 logger = logging.getLogger(__name__)
 
@@ -45,6 +47,8 @@ class Debsbom:
         spdx_namespace: tuple | None = None,  # 6 item tuple representing an URL
         cdx_serialnumber: UUID | None = None,
         timestamp: datetime | None = None,
+        snapshot_main: str | None = None,
+        snapshot_security: str | None = None,
         cdx_standard: BOM_Standard = BOM_Standard.DEFAULT,
     ):
         self.sbom_types = set(sbom_types)
@@ -66,6 +70,8 @@ class Debsbom:
 
         self.cdx_serialnumber = cdx_serialnumber
         self.timestamp = timestamp
+        self.snapshot_main = snapshot_main
+        self.snapshot_security = snapshot_security
 
         logger.info(f"Configuration: {self.__dict__}")
         self.packages: set[Package] = set()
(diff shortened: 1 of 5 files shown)


Urist-McGit:
We already have a timestamp in the BOM metadata. Correct me if I am wrong, but the difference would be that the timestamp that is already there is for creation time only, and we could still decide to use a different snapshot date which would then land here.

Also this only adds this property to the CDX format. I know there is no free-form properties in SPDX, but have you taken a look to see if we can also add it there somehow?

src/debsbom/generate/generate.py
```diff
@@ -93,6 +97,19 @@ def _import_packages(self, stream=None):
             merge_ext_states=merge_ext_states,
         )
 
+    def _extract_snapshot_from_sources_file(self):
+        if not self.snapshot_date:
+            sources_path = Path(self.root / "etc/apt/sources.list.d/0000bootstrap.list")
```
Urist-McGit: After a quick look I failed to find this file mentioned in any of the Debian documentation. Could you give me a hint where this comes from?

fmoessbauer: This is from ISAR.

fmoessbauer: I prefer to not extract the snapshot date. First, the implementation is fragile (you parse the old-style apt sources here, while upstream debian already uses deb822). Second, you can have multiple dates in there - one per repo (e.g. debian and debian-security). Providing a CLI option to set the date is IMHO fine, but also that should also be a hint.

asrinuva:
> There is also considerations how we would deal with a snapshot date in the `download` command. I would expect if such a date is in the SBOM the resolver would take that information into account.

I will try to address this one

Urist-McGit:
>If a rootfs is provided, then the snapshot date should be automatically derived from the source list—assuming the rootfs packages were built from a specific snapshot.

That is a big assumption though. There can also be multiple lists there, and we would need to identify which package was installed from where. This is a non-trivial problem to solve.

src/debsbom/generate/generate.py
Urist-McGit: Yes, I do not think the snapshot date can be easily extracted from the rootfs, so we should remove the extraction and rely solely on the input argument.

chombourger:
Having looked at your changes, you are building a list of properties. We could use --add-meta-data key1=value1 --add-meta-data key2=value2 to pass an arbitrary number of properties

Urist-McGit:
We are derailing this Pr a little bit, we should probably try to wrap this up and either merge some of this or close it in favor for additional work.

How I see it is that automatically discovering the snapshot dates is something you want, but it is hard to add. I took some time to look into how we could parse this information and annotate it properly. It should be possible and I started the implementation already, but it will be some more days before it is ready. There is also the question of where a package comes from, this would require consideration of the apt sources lists.

Adding the option to add generic data to the SBOM metadata is a feature we could add quickly. The only reservation I have is the fact that we could fragment the generated SBOMs and users would just add metadata they need manually instead of contributing or opening issues so we can solve them here for everyone.

asrinuva:
> > Should I proceed with adopting the generic approach suggested by Cedric, or do you plan to take part of these changes and merge them with your implementation?
> 
> If you feel up to the task we would appreciate if you could implement the generic metadata arguments. As I said I have some reservations, but I can definitely see the use-case. I would also say we close this PR in favor of a new one for the generic metadata argument, this also allows us to have some more discussion about it.

Sure Christof, i will close once i raise the generic metadata argument PR, thanks.

PR #154, opened 2025-12-11
feat(repack): add source and binary components filtering options

ssnaaz:


src/debsbom/commands/repack.py
```diff
@@ -48,11 +49,69 @@ def run(cls, args):
...
+                            for ref in pkg.external_references
+                            if ref.reference_type == "purl"
+                        )
+                    ]
+                    binary_ids = {pkg.spdx_id for pkg in resolver.document.packages}
+                    resolver.document.relationships = [
+                        rel
+                        for rel in resolver.document.relationships
+                        if rel.spdx_element_id in binary_ids
+                        and rel.related_spdx_element_id in binary_ids
```
Urist-McGit: This results in more or less `n**2` time complexity, as we expect to have at least one dependency for each binary package. I timed it and it turns out that is negligible compared to the other IO work done, but we should still not use a linear search here. Instead use a dictionary and perform hash lookups.

Urist-McGit:
Otherwise this looks good, we can merge it after you separated the commits. Thanks a lot again!

PR #155, opened 2025-12-11
Generate documentation as manpages as well

fmoessbauer:
As a preparation for packaging the tool, we prepare the documentation to be built as manpages as well.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -33,6 +33,12 @@ jobs:
           cd docs
           make html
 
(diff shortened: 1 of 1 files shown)


Aequitosh:
FWIW I took the liberty to test whether the manpages added here can be included in the Debian packaging stuff as well already—they successfully build and are added to the main binary package over at #157. Was able to view them on my console.

---

Very nice docs overall, btw.

`* This is a debsbom manpage.`
`* All craftsdwarfship is of the highest quality.`

PR #156, opened 2025-12-11
Add MANIFEST.in and Make Download Schema Accessible Through Module

Aequitosh:
While packaging `debsbom` for Debian, I found that some tests would fail when ran by `pybuild`, because it is assumed that [`schema-download.json`]() in a fixed path relative to the `tests/` directory (L270):
https://github.com/siemens/debsbom/blob/d8dd472ec7143cc701dde6dfbe014b00fbfb3c6a/tests/test_download.py#L268-L273

So, instead of loading [`schema-download.json`](https://github.com/siemens/debsbom/blob/d8dd472ec7143cc701dde6dfbe014b00fbfb3c6a/src/debsbom/schema/schema-download.json) in the fixture, make `debsbom.schema` a proper Python module and load it during module import. This makes the JSON accessible without requiring a specific directory structure.

I got the idea for this from the following band-aid fix I used for our internal packaging:
```diff
diff --git a/tests/test_download.py b/tests/test_download.py
index 5e7ccf2..672bbde 100644
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -259,6 +259,10 @@ def test_srcpkg_with_checksum(sdl):
 @pytest.fixture(scope="session")
 def dlschema():
     schemapath = Path(__file__).parent / "../src/debsbom/schema/schema-download.json"
+
+    if not schemapath.exists():
+        schemapath = Path(__file__).parent / "../debsbom/schema/schema-download.json"
+
     with open(schemapath) as f:
         schema = json.load(f)
     return schema
```

Since that felt a bit too dirty to upstream, I cooked up this PR instead.

Edit: Added acc8e9a5fa970b14389b5457e2dbeda29e722be6, which I was originally planning to add through a separate PR; it adds a `MANIFEST.in` file that ensures that everything under `src/` is included in the built package. This is necessary to make `schema-download.json` accessible.

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -15,6 +15,7 @@ from debsbom.download import (
 )
 from debsbom.download.download import DownloadResult, DownloadStatus
 from debsbom.resolver import PackageResolver, PackageStreamResolver
+from debsbom import schema
 from debsbom.dpkg.package import (
     BinaryPackage,
     SourcePackage,
@@ -265,14 +266,6 @@ def test_srcpkg_with_checksum(sdl):
     assert files[0].archive_name == "debian-ports"
 
 
-@pytest.fixture(scope="session")
-def dlschema():
-    schemapath = Path(__file__).parent / "../src/debsbom/schema/schema-download.json"
-    with open(schemapath) as f:
-        schema = json.load(f)
-    return schema
-
-
 @pytest.mark.parametrize(
     "dlresult",
     [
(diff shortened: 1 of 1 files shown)


Aequitosh:
Ah, nevermind, this fails in CI:
```
ERROR tests/test_download.py - FileNotFoundError: [Errno 2] No such file or directory: '/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/debsbom/schema/schema-download.json'
```

I wonder why it's included in the `.whl` for me locally, but not here... will see if I can find a fix.

Urist-McGit (approved):

PR #172, opened 2026-01-16
Prepare v0.6.0 release

Urist-McGit:
Proposed changelog:

# Improvements

- Add support for SHA512 checksums
- Improve error handling for the snapshot resolver
- add `--add-meta-data` option for `generate` command
  This option allows the user to add arbitrary metadata to the SBOM.
- Add `--source` and `--binary` options for `repack` command
  Just as in the `download` command, these options filter the packages.
- Add license information to generated SBOMs with the `--with-licenses` option
  The license information for packages is extracted from the Debian copyright files. All licenses found in there are then converted to valid SPDX license expressions and combined with `AND` to reach a final license expression for the whole source package. The approach is very conservative so we never emit wrong or incomplete license information.
- Gracefully handle incorrect source package relations
  Some packages have incorrect source package relations. Emit a warning when that happens.

# Bug fixes

- Remove spaces in VCS locator fields for SPDX SBOMs
  We were not compliant with the specification. We now drop additional details (e.g. branch, path). This information is rarely present.
- Correctly resolve virtual package dependencies
  Packages might not depend on a concrete package but instead on a virtual one. These dependencies were not correctly resolved and just skipped.
- Respect version restrictions on dependencies
  This fix goes in hand with the virtual package dependency resolution. The previous assumption was that a binary package could only be installed once and thus it had to the correct version to satisfy the dependency. This assumption is not correct if one considers virtual packages. Now always check if a package satisfies the version restrictions on a dependency.

# Documentation

- Add manpages for all commands
- Add link to documentation in README
  This was long overdue. Previously the documentation could only be reached by the badge on the github page.
- Document design decisions for the license information handling

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "debsbom"
-version = "0.5.1"
+version = "0.6.0"
 dependencies = [
   "packageurl-python>=0.16.0",
   "python-debian>=0.1.49",


fmoessbauer (approved):
A nice release. The changelog is also fine.

PR #185, opened 2026-03-03
feat: add trace_path module to get path between two components

fmoessbauer:
When working with SBOMs a common task is to analyze dependency chains, e.g. to reason why a package was installed or whom to contact regarding a dependency subtree. For that, we implement the backtrack feature that computes paths between a component and the root component.


Urist-McGit:
You forgot to add the new subcommand in the README.

PR #189, opened 2026-03-06
feat(tab-complete): mark file and directory arguments

fmoessbauer:
The current tab completion script does provide filename or directory hints for arguments that take a path (because it assumes the args just take strings). As shtab currently does not support to externally mark these arguments as paths args, we need to add this annotation to the parser.

As technically the annotations are also only needed when generating the completion script, we add shtab as an optional dependency and add helpers to mark the arguments as path args only if shtab is available. Having the paths completions significantly improves the user experience.

### DIFF
--- a/src/debsbom/cli.py
+++ b/src/debsbom/cli.py
@@ -86,6 +86,8 @@ def setup_parser():
         help="make output machine readable",
         action="store_true",
     )
+    if HAS_SHTAB_DEPS:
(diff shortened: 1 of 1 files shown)


fmoessbauer:
> does not show me any files. Does this fix the issue?

Yes.

You can test this by running:

```bash
debsbom --print-completion bash > bash_completion
.  bash_completion
```

Urist-McGit (approved):

PR #210, opened 2026-03-23
Improve and align graph export of SPDX and CycloneDX SBOMs

fmoessbauer:


src/debsbom/export/cdx.py
```diff
@@ -68,17 +72,34 @@ def _add_package(self, graph: ET.Element, p: Component):
...
+        for source in self.document.dependencies:
+            for target in source.dependencies:
+                _from = self._strip_id_prefix(source.ref)
+                _to = self._strip_id_prefix(target.ref)
+
+                # reconstruct relation types from bomref and map similar to SPDX
+                if self._arch_of_target_comp(target) == "source":
+                    reltype = "generates"
+                    _from, _to = _to, _from
+                elif self._arch_of_target_comp(source):
```
Urist-McGit: This reads a bit funny. I think we should rename `_arch_of_target_comp` to simply e.g. `_component_arch`.

src/debsbom/export/cdx.py
```diff
@@ -68,17 +72,34 @@ def _add_package(self, graph: ET.Element, p: Component):
...
+                    reltype = "package_of"
+                    _from, _to = _to, _from
+                else:
+                    reltype = "depends_on"
```
Urist-McGit, line 105: I am not quite sure I understand the logic here. What are you trying to achieve by checking if the `source` has an `arch` (or `PURL` for that matter)? If the assumption is that any binary/source package as an arch/PURL the relationships of should be reversed.

fmoessbauer: This just creates the same relations as we have in the SPDX SBOM (except for built-using). By that, we have to reverse the edges of `generates` (source-code package -> binary package)  and `package_of` (binary package -> OS meta component).

The term `source` is unfortunately ambiguous and means source in the sense of a source -> target relation, not in the sense of source code.

Urist-McGit: I do not mean reversion of the `from` and `to` direction. I mean the condition in the `elif`

Urist-McGit (approved):

PR #212, opened 2026-03-24
Add options to include recommended and suggested package dependencies

### LINKED ISSUE #209: How to model dependencies from "Recommends" and "Suggests"
A Debian package models its dependencies at multiple levels (namely `Pre-Depends`, `Depends`, `Recommends`, `Suggests`). When installing with apt, the package manager is instructed to install at a particular level (e.g. `--no-install-recommends` means it should only install the pre-depends and depends) dependencies. Unfortunately, the information why a package was installed (e.g. because the user explicitly asked for installing suggests) is not recorded. By that, we cannot strictly say if a package that is installed in the rootfs really *depends* on another package of if simply both packages are installed coincidentally.

As the dependency tree is crucial for routing package vulnerabilities to a user (another package) that can make a decision if the vulnerability affects the system, we have to precisely track these relations. For that, I propose the following (which is in-line with the behavior of apt on `purge`):

- All dependency kinds are treated as hard dependencies by simply checking if the dependent package is installed (e.g. `A ->_suggests -> B`, add a dependency from A to B if B is installed)
- Annotate the edges with the dependency type (we have to use a comment as neither CycloneDX nor SPDX supports expressing this)
- Make adding the dependencies configurable, whereby
  - `Recommends` dependencies are considered by default, except if disabled via config option
  - `Suggests` dependencies are not considered by default, except if enabled via config option
- document the behavior in the design decisions

Further, we should extend the `debsbom export` command to extract the comments and re-create the edges with their particular types. By that, the dependency tree can easily be analyzed with downstream graph tooling (like Gephi).

Xref: #183

Urist-McGit:
Add the `--recommends-deps` and `--suggested-deps` option to the generate command. These options control respectively if recommended or suggested packages are considered as dependencies in the dependency tree. These dependencies behave xactly the same as a normal dependency. In SPDX SBOMs these relationships additionally get a comment describing their exact relationship.

Per default only recommended packages are treated this way, as apt installs recommended packages, but not suggested ones in the default configuration.

Also add some minor documentation fixes I stumbled over on the way there.

Closes #209

### DIFF
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -677,6 +677,24 @@ class BinaryPackage(Package):
         """
         return self._unique_deps_for(self.all_depends)
 
+    @property
+    def unique_recommends(self) -> list[Dependency]:
+        """
+        Returns the unique recommendations without version.
+        The raw dependencies can include version specifiers, but as only a single
+        version can be installed at a time, we ignore them.
+        """
+        return self._unique_deps_for(self.recommends)
+
+    @property
+    def unique_suggests(self) -> list[Dependency]:
+        """
+        Returns the unique suggestions without version.
+        The raw dependencies can include version specifiers, but as only a single
+        version can be installed at a time, we ignore them.
+        """
+        return self._unique_deps_for(self.suggests)
+
     def merge_with(self, other: "BinaryPackage"):
         """Copy properties from other which are unset on our side. Merge lists and dicts. Or booleans."""
         super().merge_with(other)
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -576,6 +576,7 @@ class BinaryPackage(Package):
     depends: list[Dependency]
     pre_depends: list[Dependency]
     provides: list[VirtualPackage]
+    recommends: list[Dependency]
     built_using: list[Dependency]
     description: str | None
     essential: bool
@@ -595,6 +596,7 @@ class BinaryPackage(Package):
         depends: list[Dependency] = [],
         pre_depends: list[Dependency] = [],
         provides: list[VirtualPackage] = [],
+        recommends: list[Dependency] = [],
         built_using: list[Dependency] = [],
         description: str | None = None,
         essential: bool = False,
@@ -613,6 +615,7 @@ class BinaryPackage(Package):
         self.depends = depends
         self.pre_depends = pre_depends
         self.provides = provides
+        self.recommends = recommends
         self.built_using = built_using
         self.description = description
         self.essential = essential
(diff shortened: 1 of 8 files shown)


src/debsbom/dpkg/package.py
```diff
@@ -696,6 +699,10 @@ def merge_with(self, other: "BinaryPackage"):
         pre_depends.extend(x for x in other.pre_depends if x not in pre_depends)
         self.pre_depends = pre_depends
 
+        recommends = list(self.recommends)
+        recommends.extend(x for x in other.recommends if x not in recommends)
+        self.recommends = recommends
```
fmoessbauer, line 704: This is becoming a pattern. Can we please first unify the list merging and then use it?

src/debsbom/dpkg/package.py
```diff
@@ -655,22 +655,28 @@ def all_depends(self) -> Iterable[Dependency]:
...
                 seen.add(key)
                 unique.append(dep)
         return unique
 
+    @property
+    def unique_depends(self) -> list[Dependency]:
+        """
+        Returns the unique dependencies without version.
+        The raw dependencies can include version specifiers, but as only a single
+        version can be installed at a time, we ignore them.
```
fmoessbauer, line 676: A single dependency per architecture. Is this already taken care of?

tests/test_generation.py
```diff
@@ -602,3 +606,41 @@ def test_essential_required_installed(tmpdir, sbom_generator):
             ],
             "ref": "CDXRef-pytest-distro",
         } in dependencies
+
+
+def test_suggested_recommended(tmpdir, sbom_generator):
```
fmoessbauer, line 611: It should be easy to split this into a cdx and and spdx test. By that we get better test coverage, especially in the autopkgtest on trixie backports.

src/debsbom/dpkg/package.py
Urist-McGit: This function previously implemented what is now in `_unique_deps_for`. The diff displays it in a strange way I guess

tests/test_generation.py
Urist-McGit: Is this really something we want to do? This will create some overhead since we are parsing the rootfs twice, unless we also implement some clever caching of that. And it will be more of a pain to maintain, only for a backport that we will stop supporting not too far in the future.

IMO a simple patch in Salsa removing the unneeded parts of the test case would be more appropriate.

src/debsbom/dpkg/package.py
Urist-McGit: Unifying this saves one line, and the functionality of that snippet would be so basic I am not sure this is worth it.

fmoessbauer: It would save one line but 5 times.

src/debsbom/dpkg/package.py
fmoessbauer: Ah, ok.

tests/test_generation.py
fmoessbauer: Hmm... we don't need to remove anything as it simply is not executed. But in general I like the idea to be able to run parts of the testsuite if only spdx / cdx libs are available.

fmoessbauer (approved):
The remaining points are nitpicks, so I'll approve this already.

PR #214, opened 2026-04-01
feat: add sec-scan command to find vulnerabilities in SBOM packages

fmoessbauer:
We add the sec-scan command that takes an SBOM (or universal ingress) and queries the debian security tracker db for known vulnerabilities. These can be emitted in various formats, including SARIF and OpenVEX.

While security scanning is not a core domain of debsbom, adding support to query the security tracker DB is quite valuable for a quick analysis. Adding this to a dedicated tool was considered, but as debsbom already brings the infrastructure to read and process SBOMs (which apparently is quite hard to do correctly), lots of logic would have to be used via unstable interfaces or duplicated. To avoid this, we made the decision to add this feature. The only source of truth regarding vulnerability information should be the debian security tracker. There are not any plans to add support for any other security databases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ positional arguments:
     repack              repack sources and sbom
     source-merge        merge referenced source packages
     trace-path          trace path between components
+    sec-scan            check SBOM for security vulnerabilities
 
 options:
   -h, --help            show this help message and exit
@@ -65,7 +66,6 @@ At its core, this tool was designed to fulfill these SBOM generation requirement
 ### Non Goals
 
 - License and copyright text extraction from source packages
-- Real-time vulnerability database integration
 - Signing and attestation of generated artifacts
 
 ## Package Relations
--- a/src/debsbom/commands/security_scan.py
+++ b/src/debsbom/commands/security_scan.py
@@ -1,0 +1,130 @@
+# Copyright (C) 2026 Siemens
+#
+# SPDX-License-Identifier: MIT
+
+import logging
+from pathlib import Path
+
+from ..dpkg.package import filter_sources
+from ..commands.input import PkgStreamInput, SbomInput
+from ..securityscan.scanner import SecurityScanner, CveUrgency
+from ..securityscan.writer import ScanResultWriter
+
+try:
+    import requests
+
+    HAS_REQUESTS_DEP = True
+except ModuleNotFoundError:
+    HAS_REQUESTS_DEP = False
+
+
+DEBIAN_BUGTRACKER_URL = "https://bugs.debian.org/cgi-bin/bugreport.cgi"
+SECURITY_TRACKER_URL = "https://security-tracker.debian.org/tracker"
+SECURITY_DB_URL_PATH = "data/json"
+# delay path expansion to make help message and docs reproducible
+SECURITY_DB_PATH_DEFAULT = Path("~") / ".cache" / "debsbom" / "security-tracker.json"
+
+
+logger = logging.getLogger(__name__)
+
+
+class SecurityScanCmd(SbomInput, PkgStreamInput):
+    """
+    Scans packages from an SBOM for security vulnerabilities.
+    """
+
+    @classmethod
+    def download_db(cls, url: str, db_path: Path) -> None:
+        """Download the Debian security tracker JSON database."""
+        if not HAS_REQUESTS_DEP:
+            raise RuntimeError('Missing "requests" dependency')
+
+        logger.info(f"Downloading security tracker database from {url}")
+        # if the default path is used, create it
+        if db_path.expanduser() == SECURITY_DB_PATH_DEFAULT.expanduser():
+            db_path.parent.mkdir(parents=True, exist_ok=True)
+        response = requests.get(url, stream=True)
+        response.raise_for_status()
+
+        with open(db_path, "wb") as f:
+            for chunk in response.iter_content(chunk_size=8192):
+                f.write(chunk)
+        logger.info(f"Database downloaded to {db_path}")
+
+    @classmethod
+    def run(cls, args):
+        db_path = args.db.expanduser()
+        if args.update_db or not db_path.exists():
+            cls.download_db(f"{args.tracker}/{SECURITY_DB_URL_PATH}", db_path)
+
+        if cls.has_bomin(args):
+            resolver = cls.get_sbom_resolvers(args)[0]
+        else:
+            resolver = cls.get_pkgstream_resolver()
+        input_filename = Path(args.bomin) if args.bomin not in [None, "-"] else None
+
+        pkgs = list(resolver)
+        scanner = SecurityScanner(db_path, distro=args.distro)
+        vulns_it = scanner.scan(
+            filter_sources(pkgs), CveUrgency.from_string(args.min_urgency), args.filter
+        )
+        with ScanResultWriter.create(
+            "json" if args.json else args.format,
+            sdo_url=args.tracker,
+            bdo_url=DEBIAN_BUGTRACKER_URL,
+            author=args.author,
+            input_filename=input_filename,
+            packages=pkgs,
+        ) as f:
+            for v in vulns_it:
+                f.write(v)
+
+    @classmethod
+    def setup_parser(cls, parser):
+        from ..cli import arg_mark_as_dir
+
+        cls.parser_add_sbom_input_args(parser)
+        parser.add_argument(
+            "--author",
+            type=str,
+            help="author of the document (vex only)",
+        )
+        arg_mark_as_dir(
+            parser.add_argument(
+                "--db",
+                type=Path,
+                default=SECURITY_DB_PATH_DEFAULT,
+                help="path to Debian security tracker JSON database (default: %(default)s)",
+            )
+        )
+        parser.add_argument(
+            "--distro", default="trixie", help="Debian distribution to check (%(default)s)"
+        )
+        parser.add_argument(
+            "--update-db",
+            action="store_true",
+            help=f"download the security tracker database (from --tracker) "
+            "and store it at the path specified by --db",
+        )
+        parser.add_argument("--filter", type=str, help="limit search to a specific package name")
+        parser.add_argument(
+            "-f",
+            "--format",
+            choices=["text", "json", "sarif", "vex"],
+            default="text",
+            help="output format (default: %(default)s)",
+        )
+        parser.add_argument(
+            "--min-urgency",
+            type=str,
+            choices=[str(c) for c in CveUrgency],
+            default=str(CveUrgency.NOT_YET_ASSIGNED),
+            help="filter CVEs by urgency (default: %(default)s)",
+        )
+        parser.add_argument(
+            "--tracker",
+            type=str,
+            help="URL of upstream debian security tracker (default: %(default)s)",
+            default=SECURITY_TRACKER_URL,
+        )
+        return parser
--- a/src/debsbom/commands/security_scan.py
+++ b/src/debsbom/commands/security_scan.py
@@ -18,6 +18,15 @@ except ModuleNotFoundError:
     HAS_REQUESTS_DEP = False
 
 
+try:
+    from ..tracepath.walker import GraphWalker
+
+    HAS_TRACEPATH_DEPS = True
+except ModuleNotFoundError as e:
+    HAS_TRACEPATH_DEPS = False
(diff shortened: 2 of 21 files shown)


fmoessbauer:
Converting to draft, as our VEX output is not 100% OpenVEX 0.2.0 compatible.

ssnaaz:
@fmoessbauer Just gave this new feature a quick try - looks great so far.
However, I noticed that filtering for **CRITICAL** severity isn't available. Could you confirm if this was intentionally left out for now, or might have been missed?
```
debsbom sec-scan: error: argument --min-urgency: invalid choice: 'critical' (choose from high, medium, low, unimportant, end-of-life, not-yet-assigned)
```

fmoessbauer:
> @fmoessbauer Just gave this new feature a quick try - looks great so far. However, I noticed that filtering for **CRITICAL** severity isn't available. Could you confirm if this was intentionally left out for now, or might have been missed?
> 
> ```
> debsbom sec-scan: error: argument --min-urgency: invalid choice: 'critical' (choose from high, medium, low, unimportant, end-of-life, not-yet-assigned)
> ```

Is there any official documentation regarding which values are used? Ideally we would have a JSON schema. The set of values defined in debsbom are based on what I observed in the tracker data:

```
# cat ~/.cache/debsbom/security-tracker.json | jq | grep '"urgency":' | less | sort | uniq
          "urgency": "end-of-life"
          "urgency": "high"
          "urgency": "low"
          "urgency": "low",
          "urgency": "medium"
          "urgency": "not yet assigned"
          "urgency": "not yet assigned",
          "urgency": "unimportant"
          "urgency": "unimportant",
```

I'll also bring this question up on the debian security ML.

fmoessbauer:
> Converting to draft, as our VEX output is not 100% OpenVEX 0.2.0 compatible.

The output is now OpenVEX compliant and also models the source binary relations correctly in the vex document. We still need to decide if we want to keep the custom json output, or just rely on SARIF or OpenVex.

fmoessbauer:
Depends on #216

fmoessbauer:
With the recent additions, we now support (both spdx and CycloneDX):

- precisely checking which binary packages are affected by a vulnerability filed against a source package
  - based on src -> bin relation
  - based on built-using relation
- creating a SARIF report based on the tracker data that can be
  - inspected in the VS Code SARIF viewer
  - read by `vexctl`
- create a vex report for all referenced packages
  - including affected and not affected statements
  - including binary + source relation in products (including hashes)
  - including mitigation hints (e.g. update if fixed version is available)
- filter by a single package name (mostly useful for debugging)
- filter by urgency / severity

In addition, the `vexctl` tool can be used to read the generated SARIF report and filter that based on additional OpenVEX statements.

@ssnaaz It would be great if you could this PR again. Thanks!

src/debsbom/commands/security_scan.py
```diff
@@ -0,0 +1,129 @@
+# Copyright (C) 2025 Siemens
```
Urist-McGit: Nit: current year is 2026

src/debsbom/commands/security_scan.py
```diff
@@ -0,0 +1,129 @@
...
+    HAS_REQUESTS_DEP = True
+except ModuleNotFoundError:
+    HAS_REQUESTS_DEP = False
+
+
+DEBIAN_BUGTRACKER_URL = "https://bugs.debian.org/cgi-bin/bugreport.cgi"
+SECURITY_TRACKER_URL = "https://security-tracker.debian.org/tracker"
+SECURITY_DB_URL_PATH = "data/json"
+# delay path expansion to make help message and docs reproducible
+SECURITY_DB_PATH_DEFAULT = Path("~") / ".cache" / "debsbom" / "security-tracker.json"
```
Urist-McGit, line 34: We cache other things in a local folder per default. We should stay consistent and not cache parts in the local directory and parts in the home directory of the user.

fmoessbauer: Hmm... What we cache has different relations: While the download artifacts are mostly project specific, the security tracker data is global.

However, I agree that this subtle difference is hard to explain, so we better just cache in a local dir.

fmoessbauer: ... and we don't have a local directory to store the DB (in `download` we have a default one where also the download artifacts are stored).

src/debsbom/resolver/cdx.py
```diff
@@ -28,6 +36,38 @@ def document(self):
...
+                # self is binary, other is source
+                if pkg_other.is_source():
+                    if pkg_self.source:
+                        # as we cannot distinguish between binary->src package and built-using,
+                        # we just add the remaining packages into built-using
+                        pkg_self.built_using.append(
+                            Dependency(pkg_other.name, version=("=", pkg_other.version))
+                        )
+                        continue
+                    pkg_self.source = Dependency(pkg_other.name, version=("=", pkg_other.version))
```
Urist-McGit: So the assumption here is that the actual source package for our binary package is coming first in the dependency list? I guess this works because of our implementation, but maybe we should explicitly state here that we do it this way and it just an assumption.

Or we could use a simple heuristic that would work for most packages: if the name of the binary and source package are the same, assume they belong together. And otherwise just assume the first package.

src/debsbom/commands/security_scan.py
Urist-McGit: Thinking about it, I guess it is fine to store it in the home directory.  `~/.cache` is a well-known path, and the location can be changed.

src/debsbom/commands/security_scan.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @classmethod
+    def download_db(cls, url: str, db_path: Path) -> None:
+        """Download the Debian security tracker JSON database."""
+        if not HAS_REQUESTS_DEP:
+            raise RuntimeError('Missing "requests" dependency')
+
+        logger.info(f"Downloading security tracker database from {url}")
+        if db_path.is_relative_to(SECURITY_DB_PATH_DEFAULT.expanduser().parent):
+            db_path.parent.mkdir(parents=True, exist_ok=True)
```
Urist-McGit, line 54: This is a bit convoluted. What you want to do is create the directory structure only if the default path is used? In that case it might be easier to instead pass `None` as argument and check for that, or directly check for the default path.

src/debsbom/commands/security_scan.py
```diff
@@ -0,0 +1,129 @@
...
+
+        with open(db_path, "wb") as f:
+            for chunk in response.iter_content(chunk_size=8192):
+                f.write(chunk)
+        logger.info(f"Database downloaded to {db_path}")
+
+    @classmethod
+    def run(cls, args):
+        db_path = args.db.expanduser()
+        if args.download_db or not db_path.exists():
```
Urist-McGit: This can lead to unexpected behavior. If there is an option that is named `--download-db` I only expect it to be downloaded when that is set. We would download it also if the path does not exist. It would be better to print an error here, stating that the db is not available and to download it with `--download-db`.

src/debsbom/commands/security_scan.py
```diff
@@ -0,0 +1,129 @@
...
+            "--author",
+            type=str,
+            help="author of the document (vex only)",
+        )
+        arg_mark_as_dir(
+            parser.add_argument(
+                "--db",
+                type=Path,
+                default=SECURITY_DB_PATH_DEFAULT,
+                help="path to Debian security tracker JSON database " "(%(default)s)",
```
Urist-McGit: Is there any reason you concat two strings here?

You should also state that the value in brackets is the default, so `(default: %(default)s)`. Otherwise people might be confused, as e.g. for `--author` the value in brackets is not the default. This also goes all other arguments where there is a default here.

src/debsbom/securityscan/scanner.py
```diff
@@ -0,0 +1,119 @@
...
+        return cls[s.upper().replace(" ", "_").replace("-", "_")]
+
+    def __str__(self) -> str:
+        return self.name.lower().replace("_", "-")
+
+
+class CveStatus(IntEnum):
+    RESOLVED = (0,)
+    UNDETERMINED = (1,)
+    OPEN = (2,)
```
Urist-McGit: You use tuples here instead of simply assigning the numbers. What is the reason?

src/debsbom/securityscan/writer.py
```diff
@@ -0,0 +1,338 @@
...
+                "ruleId": rule_id,
+                "level": "warning" if v.urgency == CveUrgency.HIGH else "note",
+                "message": {
+                    "text": f"The SBOM reports {bin_pkg.name} at version {bin_pkg.version} which is a vulnerable deb package affected by {v.cve}",
+                },
+                "locations": [
+                    {
+                        "physicalLocation": {
+                            "artifactLocation": {
+                                "uri": f"file://{self.path}" if self.path else "<stdin>",
```
Urist-McGit: A `<stdin>` location is not at all useful, is it? Are the locations mandatory? Otherwise we should just leave the location field out if we get the data from stdin.

src/debsbom/securityscan/writer.py
```diff
@@ -0,0 +1,338 @@
...
+                "@id": f"{self.sdo_url}/{v.cve}",
+                "name": v.cve,
+            },
+            "products": products,
+            "status": status,
+        }
+        if v.description:
+            vex["vulnerability"]["description"] = v.description
+        if status == "not_affected":
+            vex["justification"] = "inline_mitigations_already_exist"
```
Urist-McGit, line 349: I know we have to put something here, but is this always correct? IMO the `vulnerable_code...` justifications might also fit here, but it depends on how Debian actually reports these vulnerabilities

README.md
```diff
@@ -65,7 +66,6 @@ At its core, this tool was designed to fulfill these SBOM generation requirement
 ### Non Goals
 
 - License and copyright text extraction from source packages
-- Real-time vulnerability database integration
```
Urist-McGit, line 68: Mandatory feature creep reminder ;)

Urist-McGit (changes requested):

src/debsbom/commands/security_scan.py
fmoessbauer: > This is a bit convoluted. What you want to do is create the directory structure only if the default path is used?

Then let's just check for the standard path.

src/debsbom/commands/security_scan.py
fmoessbauer: How about renaming this to `--update-db` and keep the current behavior? Especially for CI systems we would like to keep the single-call behavior.

src/debsbom/securityscan/scanner.py
fmoessbauer: For standard enums you can avoid the auto-conversion, but as this is an `intEnum` we better just use ints.

src/debsbom/securityscan/writer.py
fmoessbauer: Downstream viewers expect that to open the document when clicking on the result. However I agree that we can leave out the `physicalLocation` if the path is none. Then we can also avoid the whole `<stdin>` part.

src/debsbom/securityscan/writer.py
fmoessbauer: Right, I'll add a comment. There is a discussion on the security tracker regarding this [1][2]. Once resolved, we can improve the implementation.

- [1] https://salsa.debian.org/security-tracker-team/security-tracker/-/issues/41
- [2] https://salsa.debian.org/security-tracker-team/security-tracker/-/issues/38

README.md
fmoessbauer: Thanks for the reminder :)

src/debsbom/resolver/cdx.py
fmoessbauer: > Or we could use a simple heuristic that would work for most packages: if the name of the binary and source package are the same, assume they belong together. And otherwise just assume the first package.

This is a good idea: We add the first package to source so we can be sure we have a source. Then, we iterate and check if package names are matching and if so, move the current source to built_using and replace the source with the processed one.

src/debsbom/securityscan/writer.py
Urist-McGit: Sounds good. In the meanwhile we can leave it like that then, as that is how Debian reports them.

Urist-McGit (approved):

PR #216, opened 2026-04-08
A couple of minor fixes

fmoessbauer:


Urist-McGit:
Good catch on the default arguments! This is a Python footgun I always forget about

Urist-McGit (approved):

PR #224, opened 2026-04-24
Add options to control product behavior for security scans in VEX format

Urist-McGit:
In the vex format each vulnerability statement has a product attached to it. Add the --default-product option that contols the default behavior.
The previous behavior is achieved with the component choice, which fills in the each affected source package as the product. The other choice is distribution, which sets a global product for each statement with the root componentn name.

Also add the --product option to manually specify a global product for
the vex document.

### DIFF
--- a/src/debsbom/commands/security_scan.py
+++ b/src/debsbom/commands/security_scan.py
@@ -116,7 +116,7 @@ class SecurityScanCmd(SbomInput, PkgStreamInput):
             )
         )
         parser.add_argument(
-            "--distro", default="trixie", help="Debian distribution to check (%(default)s)"
+            "--distro", default="trixie", help="Debian distribution to check (default: %(default)s)"
         )
         parser.add_argument(
             "--update-db",
--- a/src/debsbom/commands/security_scan.py
+++ b/src/debsbom/commands/security_scan.py
@@ -105,7 +105,7 @@ class SecurityScanCmd(SbomInput, PkgStreamInput):
         parser.add_argument(
             "--author",
             type=str,
-            help="author of the document (vex only)",
+            help="author of the document (-f vex only)",
         )
         arg_mark_as_dir(
             parser.add_argument(
--- a/src/debsbom/commands/security_scan.py
+++ b/src/debsbom/commands/security_scan.py
@@ -80,6 +80,13 @@ class SecurityScanCmd(SbomInput, PkgStreamInput):
             resolver = cls.get_pkgstream_resolver()
         input_filename = Path(args.bomin) if args.bomin not in [None, "-"] else None
 
+        if args.product:
+            product = args.product
+        elif args.default_product == "distribution":
+            product = resolver.root_component_name()
+        else:
+            product = None
+
         pkgs = list(resolver)
         scanner = SecurityScanner(db_path, distro=args.distro)
         vulns_it = scanner.scan(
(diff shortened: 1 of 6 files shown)


src/debsbom/commands/security_scan.py
```diff
@@ -62,6 +62,9 @@ def download_db(cls, url: str, db_path: Path) -> None:
 
     @classmethod
     def run(cls, args):
+        if args.format != "vex" and args.product:
```
fmoessbauer: I would rather silently ignore this to behave consistently with `--author`, which is also vex only.

src/debsbom/resolver/cdx.py
```diff
@@ -80,6 +80,10 @@ def _resolve_relations(self) -> None:
                         Dependency(pkg_other.name, version=("=", pkg_other.version))
                     )
 
+    def root_component_name(self) -> str | None:
+        """Return the name of the root component."""
+        return self._document.metadata.component.name
```
fmoessbauer: This is an optional field, whereby even the component is already optional. This probably currently fails if no `metadata.component` is set.

src/debsbom/securityscan/writer.py
```diff
@@ -303,42 +309,48 @@ def _get_status(v: CveEntry, affected):
...
-                for algo, value in r.package.checksums.items()
-                if algo.name in _CHECKSUM_TO_VEX_HASH
-            }
-        products = [product]
-        for bin_pkg in self.affected_binaries(r.package):
-            bin_purl = str(bin_pkg.purl())
-            bin_product = {
-                "@id": bin_purl,
+        if not self.product:
+            # if we have a product we do not need to emit information per component
```
fmoessbauer, line 313: Here we don't have a product, so what's the comment about?

src/debsbom/securityscan/writer.py
```diff
@@ -303,42 +309,48 @@ def _get_status(v: CveEntry, affected):
...
+            product = {
+                "@id": purl,
```
fmoessbauer, line 315: Are we sure we always have a PURL for the product?

src/debsbom/resolver/cdx.py
Urist-McGit: It would fail with an exception. For SPDX we just return None in that case, so we can do the same here

src/debsbom/securityscan/writer.py
Urist-McGit: It should probably a line up and indentation up.

src/debsbom/securityscan/writer.py
Urist-McGit: I did not touch this logic, I just had to indent it since it is conditional now. But yeah, in general we can not rely on there being a PURL. In that case it is probably the best to at least give the package name

src/debsbom/securityscan/writer.py
fmoessbauer: Ah... then that should be fine.

src/debsbom/securityscan/writer.py
fmoessbauer: For our debian packages we have - by design - a purl, as the SBOM reader checks for debian purls. However, for the generic product (or distro component), we usually don't have one.

With the patch rendering, I'm unsure if the purl can also refer to a distro component (which does not work), or just to debian components (which is fine).

Urist-McGit: It refers just to Debian components. The distro component is not set here as this is the per component product, the distro component is set globally in the encapsulating document.

fmoessbauer (approved):
Thanks!

fmoessbauer:
@Urist-McGit My merge might have been a bit too fast. We are now just moving the product up the hierarchy, which is fine, but we can do better. If we have the enclosing product, we still want to add each component with a CVE as a subcomponent. By that, we can maintain the relation between the two.

PR #225, opened 2026-04-27
feat(sec-scan-vex): add affected components to subcomponents

fmoessbauer:
When linking the vex statements to a product, we also add the affected components (the ones which are vulnerable) to subcomponents. This enables downstream users to get precise information in which component a vulnerability is present.

### DIFF
--- a/src/debsbom/securityscan/writer.py
+++ b/src/debsbom/securityscan/writer.py
@@ -284,7 +284,7 @@ class ScanResultVexWriter(ScanResultWriter):
         self.out.write("\n")
 
     def _create_skeleton(self) -> dict:
-        frame = {
+        return {
             "@context": VEX_CONTEXT,
             "@id": VEX_SCHEMA_ID,
             "author": self.author,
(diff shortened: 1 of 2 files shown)


src/debsbom/securityscan/writer.py
```diff
@@ -293,8 +293,6 @@ def _create_skeleton(self) -> dict:
             "tooling": "debsbom {}".format(version("debsbom")),
             "statements": [],
         }
-        if self.product:
-            frame["product"] = {"@id": self.product}
         return frame
```
Urist-McGit: You can directly return the `frame` instead of saving it in a local variable and then returning it.

tests/test_security_scan.py
```diff
@@ -379,5 +379,9 @@ def test_vex_with_product(scanner):
             writer.write(r)
 
     data = json.loads(buf.getvalue())
-    assert data["statements"][0].get("products") is None
-    assert data["product"]["@id"] == "Product"
+    jsonschema.validate(instance=data, schema=vex_schema)
```
Urist-McGit, line 382: Nit: this mixes the json schema validation with the changed test logic.

fmoessbauer: Moved to separate commit.

src/debsbom/securityscan/writer.py
fmoessbauer: Makes sense. Changed.

Urist-McGit (approved):

PR #229, opened 2026-05-13
Include Static-Built-Using in dependencies

Urist-McGit:
Static-Built-Using relationships are very similar to Built-Using relationships, but there is an important difference. The `man 5 deb-control` states:
```
Built-Using: package-list
           This dependency field lists extra source packages that were used during the build of this binary package, for license compliance purposes.  This is an indication to the archive  maintenance  software  that  these
           extra  source  packages must be kept whilst this binary package is maintained.  This field must be a comma-separated list of source package names with strict ‘=’ version relationships enclosed within parenthesis.
           Note that the archive maintenance software is likely to refuse to accept an upload which declares a Built-Using relationship which cannot be satisfied within the archive.

Static-Built-Using: package-list
           This dependency field lists extra source packages that were used during the build of this binary package, for static building purposes (for example linking against static  libraries,  builds  for  source-centered
           languages  such  as Go or Rust, usage of header-only C/C++ libraries, injecting data blobs into code, etc.).  This is useful to track whether this package might need to be rebuilt when source packages listed here
           have been updated, for example due to security updates.  This field must be a comma-separated list of source package names with strict ‘=’ version relationships enclosed within parenthesis.
```
TLDR: Built-Using is primarily for license purposes. Static-Built-Using for rebuilding and security purposes.

Either way we need to include both of them in our dependency graph. Currently Static-Built-Using is mostly used in the Rust and Go ecosystems.

### DIFF
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -395,6 +395,9 @@ class Package(ABC):
             # a more complete source package we discovered via other mechanisms (e.g. apt cache).
             logger.debug(f"Found built-using source package: '{bu.name}@{bu.version[1]}'")
             yield SourcePackage(bu.name, bu.version[1])
+        for sbu in pkg.static_built_using:
+            logger.debug(f"Found static-built-using source package: '{sbu.name}@{sbu.version[1]}'")
+            yield SourcePackage(sbu.name, sbu.version[1])
         if add_pkg:
             yield pkg
 
@@ -579,6 +582,7 @@ class BinaryPackage(Package):
     recommends: list[Dependency]
     suggests: list[Dependency]
     built_using: list[Dependency]
+    static_built_using: list[Dependency]
     description: str | None
     essential: bool
     priority: DebianPriority | None
@@ -600,6 +604,7 @@ class BinaryPackage(Package):
         recommends: list[Dependency] | None = None,
         suggests: list[Dependency] | None = None,
         built_using: list[Dependency] | None = None,
+        static_built_using: list[Dependency] | None = None,
         description: str | None = None,
         essential: bool = False,
         priority: DebianPriority | None = None,
@@ -620,6 +625,7 @@ class BinaryPackage(Package):
         self.recommends = recommends or []
         self.suggests = suggests or []
         self.built_using = built_using or []
+        self.static_built_using = static_built_using or []
         self.description = description
         self.essential = essential
         self.priority = priority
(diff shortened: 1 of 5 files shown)


Urist-McGit:
`python-debian`  also just merged my PR to add support for Static-Built-Using there: https://salsa.debian.org/python-debian-team/python-debian/-/merge_requests/158/commits

src/debsbom/dpkg/package.py
```diff
@@ -813,9 +825,21 @@ def from_deb822(cls, package) -> "BinaryPackage":
...
-        sdepends = Dependency.from_pkg_relations(s_built_using, is_source=True)
+        # Built-Using relationships are primarily used for license purposes
+        built_using = package.relations["built-using"] or []
+        budepends = Dependency.from_pkg_relations(built_using, is_source=True)
+
+        # statically linked dependencies are modeled with Static-Built-Using
+        #
+        # unlike all other relationships this is not yet included in the relations field as
+        # of python-debian 1.1.0 so we have to parse it by hand
+        s_built_using = package.get("Static-Built-Using")
```
fmoessbauer, line 836: What we get depends on the version of `python-debian`, right. If the version is too old, we simply get nothing instead of an error that can be caught. By that, a user of our SBOM cannot know if static-built-using relations could be processed during generation (based on the debsbom meta node in the SBOM). This is bad.

We at least need to issue a warning on generation, if the `python-debian` version is too old AND add a flag to the debsbom meta node (`e.g. <debsbom-version>+sbu`) or similar.

src/debsbom/generate/spdx.py
```diff
@@ -370,6 +370,18 @@ def spdx_bom(
                 logger.debug(f"Created built-using relationship: {relationship}")
                 relationships.append(relationship)
 
+        if package.static_built_using:
+            for dep in package.static_built_using:
```
fmoessbauer: No need for the if, as `package.static_built_using` is a list. Same applies to the `package.built_using` btw.

src/debsbom/dpkg/package.py
Urist-McGit: I think there is a misunderstanding about the comment here. The Static-Built-Using relationship can always be parsed, no matter the `python-debian` version. If you look at the other relations we can access them with the `relations` property, which we can not do for Static-Built-Using. That is why I need to parse them in two steps. With the merged PR in `python-debian` we would simply get a diff like this:
```
s_built_using = package.relations["static-built-using"] or []
sbudepends = Dependency.from_pkg_relations(s_built_using, is_source=True)
```

src/debsbom/dpkg/package.py
```diff
@@ -738,6 +744,12 @@ def merge_with(self, other: "BinaryPackage"):
         built_using.extend(x for x in other.built_using if x not in built_using)
         self.built_using = built_using
 
+        static_built_using = list(self.static_built_using)
```
fmoessbauer, line 747: Maybe it’s time to replace the open-coded list merging with a function.

fmoessbauer (approved):
Apart from the list merging, this is fine.

PR #231, opened 2026-05-28
Prepare v0.8.0 release

Urist-McGit:
Proposed changelog:

# Improvements

- Add new `sec-scan` command to find vulnerabilities in SBOM packages
  The command takes an SBOM, or a list of packages with universal ingress, and queries the debian security tracker for known vulnerabilities. Various standardized output formats, including SARIF and OpenVEX, are supported.
- Emit `Static-Built-Using` dependencies for packages
  These dependencies are similar to Built-Using dependencies, but there is a minor difference: Static-Built-Using is primarily intended for package rebuilding and cybersecurity purposes, while the old Built-Using is primarily for license obligation reasons. Since we want to cover the security related use-cases too we need to track it and emit it in generated SBOMs.

# Bug fixes

- Fix a crash with packageurl-python versions <=0.17.0 when resolving packages
  We allow packageurl-python versions >=0.16.0 and want to keep that for Debian compatibility reasons.


fmoessbauer:
While improving the autopkgtest on the Debian packaging, I noticed more dependency glitches.

fmoessbauer (approved):

PR #234, opened 2026-06-29
Add --package option to the filter command

Urist-McGit:
Add a new --package option to the filter command that allows extraction of the subgraph of the selected package in an SBOM. It is compatible with the --binaries and --sources options.

~Marked as Draft as tests are still missing.~


pyproject.toml
```diff
@@ -47,7 +47,7 @@ mindev = [
   "beartype>=0.20",
   "jsonschema>=3.2",
   "lz4",
-  "debsbom[tracepath]",
+  "debsbom[graph]",
```
fmoessbauer, line 50: Technically this is a breaking change, however I don't think anybody cares about this.
Nonetheless, we should at least mention it in the next release notes.

docs/source/api.rst
```diff
@@ -69,10 +69,10 @@ Graph Exporting
 .. automodule:: debsbom.export.graphml
    :members:
 
-Dependency Trace Paths
+Graph Traversal
 ----------------------
```
fmoessbauer: Fix the length of the underline.

pyproject.toml
Urist-McGit: Yes, this is definitely a breaking change. In the next changelog I will introduce a breaking change section where we can document any breaking changes. This will also help us prepare for the eventual 1.0 release where we will have to provide some API guarantees.

docs/source/examples.rst
```diff
@@ -243,6 +243,31 @@ You can also pass SBOMs via stdin, but you also have to pass the SBOM type in th
...
+    debsbom filter --binaries sbom.spdx.json filtered.spdx.json
+    debsbom filter --sources sbom.spdx.json filtered.spdx.json
+
+You can also get the subgraph for a specific package:
+
+.. code-block:: bash
+
+    debsbom filter --package debsbom sbom.spdx.json package-filtered.spdx.json
+
+Combining multiple filter works as expected, this would give you all source packages ``debsbom`` transitively depends on:
```
fmoessbauer: Do we somewhere document that the filters are and'ed?

docs/source/examples.rst
```diff
@@ -243,6 +243,31 @@ You can also pass SBOMs via stdin, but you also have to pass the SBOM type in th
...
+.. code-block:: bash
+
+    debsbom filter --binaries sbom.spdx.json filtered.spdx.json
+    debsbom filter --sources sbom.spdx.json filtered.spdx.json
+
+You can also get the subgraph for a specific package:
+
+.. code-block:: bash
+
+    debsbom filter --package debsbom sbom.spdx.json package-filtered.spdx.json
```
fmoessbauer, line 273: Can we please add a smoke test and example for the pipe case as well:

```bash
debsbom generate -t spdx -o - | debsbom filter --package debsbom -
```

docs/source/examples.rst
Urist-McGit: We do not. It should probably be done in the filter subcommand docs, but we should have it in the example here too.

docs/source/examples.rst
Urist-McGit: A working command would be
```
debsbom generate -t spdx -o - | debsbom filter --package debsbom --sbom-type spdx - filtered.spdx.json
```

I will add a smoke test and this line to the example.

src/debsbom/commands/input.py
```diff
@@ -229,6 +229,23 @@ def parser_add_source_binary_args(cls, parser):
...
+            from ..filter.cdx import CdxSbomFilter
+
+            if sources:
+                CdxSbomFilter.source_pkgs(resolver.document)
+            elif binaries:
+                CdxSbomFilter.binary_pkgs(resolver.document)
+        elif resolver.sbom_type() == SBOMType.SPDX:
+            from ..filter.spdx import SpdxSbomFilter
+
+            if sources:
```
fmoessbauer: This is a behavior change: Previously `--sources --binaries` was a no-op. Now it is `--sources`. It would be better to make the options either-or.

tests/test_filter.py
```diff
@@ -109,3 +130,204 @@ def test_cdx_filter_binaries(tmpdir):
...
+        "relatedSpdxElement": "SPDXRef-python3-packageurl-all",
+    } in relationships
+    assert {
+        "spdxElementId": "SPDXRef-py-serializable-2.0.0-2-srcpkg",
+        "relationshipType": "GENERATES",
+        "relatedSpdxElement": "SPDXRef-python3-py-serializable-all",
+    } in relationships
+
+
+def test_cdx_filter_binaries(tmpdir):
```
fmoessbauer: This shadows the existing test at line 114. Please rename.

src/debsbom/commands/filter.py
```diff
@@ -34,3 +89,9 @@ def setup_parser(cls, parser):
             help="validate generated SBOM (only for SPDX)",
             action="store_true",
         )
+        parser.add_argument(
+            "-p",
+            "--package",
+            type=str,
+            help="filter the SBOM by only including the package and its dependent subgraph",
```
fmoessbauer: Nit: Isn't that the `dependency subgraph`? I prefer to stick to the term dependencies, as dependants implies the other direction of dependencies.

src/debsbom/commands/filter.py
```diff
@@ -2,19 +2,74 @@
...
+                filter(lambda c: c.is_binary(), resolver.component_by_name(args.package))
+            )
+            if not len(candidates):
+                raise ValueError(f"no binary package can be found for name {args.package}")
+            else:
+                chosen = candidates[0]
+                purl = chosen.purl()
+                if len(candidates) > 1:
+                    logger.warning(
+                        "multiple binary packages match given package name, choosing '{purl}'"
```
fmoessbauer: The f-string is missing. We always log `{purl}`.

src/debsbom/graph/walker.py
```diff
@@ -102,3 +102,8 @@ def all_shortest(self, source: PackageURL) -> Iterable[list[PackageRepr]]:
     def all_simple(self, source: PackageURL) -> Iterable[list[PackageRepr]]:
         """Return all non-cyclic path from source to root. Abstract method."""
         raise NotImplementedError()
+
+    @abstractmethod
+    def descendants(self, source: PackageURL) -> Iterable[list[PackageRepr]]:
```
fmoessbauer: Isn't the return type `Iterable[PackageRepr]`?

src/debsbom/commands/filter.py
```diff
@@ -2,19 +2,74 @@
...
 
 class FilterCmd(SbomInput, SourceBinaryInput):
-    """Filter SBOM by sources or binaries."""
+    """Filter SBOMs."""
 
     @classmethod
     def run(cls, args):
         resolvers = cls.get_sbom_resolvers(args)
 
+        def get_package_repr(sbom_type: SBOMType) -> (PackageRepr, str):
```
fmoessbauer: Isn't the correct type `tuple[PackageRepr, str]`? I'm actually unsure.

tests/test_filter.py
```diff
@@ -17,6 +20,24 @@
 # > x11vnc 0.9.16-9 amd64
 # > EOF
 
+# The package-fitler SBOM was generated using this package list:
```
fmoessbauer: Typo. `package-filter`.

tests/test_filter.py
```diff
@@ -109,3 +130,204 @@ def test_cdx_filter_binaries(tmpdir):
...
+    dependencies = cdx_json["dependencies"]
+
+    assert all([unquote(component["purl"]) in EXPECTED_PURLS for component in components])
+    assert all(
+        [
+            unquote(component["purl"]) != "pkg:deb/debian/dash@0.5.12-12?arch=amd64"
+            for component in components
+        ]
+    )
+    assert all(
```
fmoessbauer, line 289: duplicated test?

src/debsbom/commands/filter.py
Urist-McGit: I think `tuple[...]` is the correct way

src/debsbom/commands/input.py
Urist-McGit: ~I introduced a mutually exclusive args group to handle this case now~

Nevermind, this is the wrong approach. I simply kept the original check if both are enabled as is.

tests/test_filter.py
Urist-McGit: Copy-and-paste error, fixed now

fmoessbauer (approved):

PR #235, opened 2026-06-30
fix(download): fix potential path traversal with remote filenames

Urist-McGit:
Properly sanitize the paths returned by any download resolver and throw an error if such a path is detected, as any such path does not occur in a valid response and is certainly malicious.

The bug is not security relevant as the the debsbom provided implementation only connects to snapshot.debian.org. This connection is properly secured by TLS and the server trustworthy. For custom resolver plugins the user is expected to verify both the validity of the plugin and the trustworthiness of the accessed servers.

### DIFF
--- a/src/debsbom/download/download.py
+++ b/src/debsbom/download/download.py
@@ -81,9 +81,18 @@ class PackageDownloader:
 
     def _target_path(self, pkg: package.Package, f: RemoteFile):
         if pkg.is_source():
-            return Path(self.sources_dir / f.archive_name / f.filename)
+            base = self.sources_dir
         else:
-            return Path(self.binaries_dir / f.archive_name / f.filename)
+            base = self.binaries_dir
+        base = base.resolve()
+        target = (base / f.archive_name / f.filename).resolve()
+        if not target.is_relative_to(base):
+            purl = pkg.purl()
+            raise ValueError(
+                f"{purl}: malicious path in remote file detected: '{f.archive_name}/{f.filename}'"
+            )
+        else:
+            return target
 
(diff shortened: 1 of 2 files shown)


src/debsbom/download/download.py
```diff
@@ -81,9 +81,18 @@ def __init__(
 
     def _target_path(self, pkg: package.Package, f: RemoteFile):
         if pkg.is_source():
-            return Path(self.sources_dir / f.archive_name / f.filename)
+            base = self.sources_dir
         else:
-            return Path(self.binaries_dir / f.archive_name / f.filename)
+            base = self.binaries_dir
+        base = base.resolve()
+        target = Path(base / f.archive_name / f.filename).resolve()
```
fmoessbauer: As base is already a Path instance, there is no need for the Path constructor.

src/debsbom/download/download.py
```diff
@@ -81,9 +81,18 @@ def __init__(
...
+        if not (base in target.parents or base == target):
```
fmoessbauer: Better use `is_relative_to`?

Urist-McGit: I am not sure this makes a difference. Looking at the implementation, `is_relative_to` tries to build a relative path and if that fails the path is considered not relative. Checking if the base dir is a parent seems like the more straightforward approach.

fmoessbauer: Please use `is_relative_to` instead of the open-coded variant. The mentioned code is (at least for me) hard to read.

fmoessbauer (approved):

PR #237, opened 2026-07-03
docs(container): add example using skopeo and umoci

michaeladler:
This is typically more practical in CI-like environments than having to properly setup Podman (or Docker).


Urist-McGit:
Looks good, thanks a lot!

Urist-McGit (approved):

PR #241, opened 2026-07-13
fix(container): add Debian patch tools to container

### LINKED ISSUE #240: Prebuilt container image is missing `dpkg-dev`, breaking `source-merge`/`repack --apply-patches`
## Summary

The prebuilt `ghcr.io/siemens/debsbom:latest` (v0.8.1) image does not include `dpkg-dev`, so any command that applies Debian patches (`source-merge --apply-patches` and `repack --apply-patches` ) fail immediately with:

```
debsbom: error: 'dpkg-source' from the 'dpkg-dev' package is missing.
```

As a result the documented license-clearing workflow cannot be run end-to-end inside the container: `generate` and `download` work, but the merge/repack step that combines and patches the downloaded sources does not.

## Affected image / version

- Image: `ghcr.io/siemens/debsbom:latest` (`debsbom 0.8.1`)
- `Dockerfile` on `main` (as of `ce377ed`)

## Environment check

```
$ docker run --rm ghcr.io/siemens/debsbom:latest \
    sh -lc 'for t in dpkg-source patch xz bzip2; do command -v $t >/dev/null || echo "$t: missing"; done'
dpkg-source: missing
patch: missing
xz: missing
bzip2: missing
```

The image ships `tar`, `gzip`, and `lz4`, but not `dpkg-source` (from `dpkg-dev`), `patch`, `xz`, or `bzip2`.

## Steps to reproduce

Minimal and fully containerized, mirroring the `download` container example from the docs (no rootfs required):

```bash
mkdir -p downloads

# 1) Download a source package inside the container -> works
echo "guestfs-tools 1.52.3-1 source" | \
  docker run --rm -i -v "$(pwd)/downloads:/mnt/downloads" \
    ghcr.io/siemens/debsbom:latest \
    debsbom download --outdir /mnt/downloads --sources

# 2) Merge and apply Debian patches inside the container -> fails
echo "guestfs-tools 1.52.3-1 source" | \
  docker run --rm -i -v "$(pwd)/downloads:/mnt/downloads" \
    ghcr.io/siemens/debsbom:latest \
    debsbom source-merge --pkgdir /mnt/downloads/sources \
                         --outdir /mnt/downloads/sources \
                         --apply-patches
```

`repack --apply-patches` fails identically, since it merges via the same `dpkg-source` path.

## Actual behavior

```
ERROR:debsbom.cli:'dpkg-source' from the 'dpkg-dev' package is missing.
debsbom: error: 'dpkg-source' from the 'dpkg-dev' package is missing.
```

The missing-dependency guard reports the problem correctly, but the required tool is simply not shipped in the image.

## Expected behavior

The prebuilt image should be able to run the full documented license-clearing workflow (`generate` -> `download --sources` -> `source-merge`/`repack --apply-patches`) without users having to build a derived image or install extra packages.

## Root cause

The image installs the Python runtime dependencies plus `lz4`, but not `dpkg-dev`. The requirement is already documented for the merge step (`docs/source/examples.rst`, "Merge Source Packages"):

> Internally, the `dpkg-source` command from the `dpkg-dev` package is used to perform the merge.

So the dependency is known; it is just not provisioned in the container. The tools `dpkg-source` needs are also absent: `patch` (to apply `3.0 (quilt)` patches) and the `xz`/`bzip2` decompressors that `dpkg-source` exec's to unpack `.orig.tar.xz`/`.orig.tar.bz2` upstream tarballs (the shared libraries alone are not enough; the CLI binaries must be on `PATH`).

## Proposed fix

Append `dpkg-dev`, `xz-utils`, and `bzip2` to the existing `apt-get install` block in the `Dockerfile`:

```dockerfile
    apt-get install --no-install-recommends -y \
        python3-apt ...
        ... 
        dpkg-dev \
        xz-utils \
        bzip2 && \
    rm -rf /var/log/* /tmp/* /var/tmp/* /var/cache/ldconfig/aux-cache
```

- `dpkg-dev` provides `dpkg-source` and pulls in `patch`; 
- `xz-utils` and `bzip2` provide the CLI decompressors `dpkg-source` calls for `.tar.xz`/`.tar.bz2` sources. 

They are `dpkg-dev` dependencies, listed explicitly because the image builds with `--no-install-recommends`. This mirrors the earlier fix for the missing `lz4` decompressor (#150).

### On image size / a separate image

Pulling in `dpkg-dev` also brings its dependencies (`perl`, `make`, `binutils`, ...), which grows an image that most `generate`/`download` users never need. 

Two options:

- **Single image (simplest):** accept the modest size increase and keep one image that runs the whole documented workflow. Consistent with the current single-image model and the `lz4` precedent (#150).
- **Dedicated image variant:** publish a separate tag (e.g. `ghcr.io/siemens/debsbom:<ver>-repack` or `-devel`), built from the same base with the extra tooling, so the default image stays slim and only the repack/source-merge/license-clearing use case pulls the heavier one. Cost: an extra CI build + tag, some docs, and a risk of drift between variants.

## Willing to contribute

Happy to open a PR for whichever direction you prefer.

hakandilek:
This change adds the missing dpkg-dev and packaging tools, needed by source-merge and repack --apply-patches

Fixes #240

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,10 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
         python3-networkx python3-ply \
         python3-rdflib python3-semantic-version python3-uritools python3-xmltodict \
         python3-yaml python3-zstandard python3-requests \
-        lz4 && \
+        lz4 \
+        dpkg-dev \
+        xz-utils \
+        bzip2 && \
     rm -rf /var/log/* /tmp/* /var/tmp/* /var/cache/ldconfig/aux-cache
 
 # install debsbom in a reproducible way


hakandilek:
@Urist-McGit could you please check this one out?

Urist-McGit (approved):

Issue #153, opened 2025-12-11
Inclusion of License Information


Urist-McGit:
Right now it is declared as a non-goal, but we might want to reconsider that.

We currently do not include license information for the following reasons:

1. For ~20% of packages license information is not machine-readable, limiting how much information we can provide
2. Packages seldomly carry only a single license, a representation per file is needed which requires some implementation effort
3. There are some legal considerations which might reduce the usefulness of the license information

We recently had some discussion with internal and Debian folks, which brought up some reasons why we might want to add it:

1. Even incomplete license information has value
2. Debian wants to actively improve the license situation. If we start incorporating/using it we could improve the situation in Debian over time, in turn improving our SBOMs
3. It should be possible to be 100% correct about any license information we provide
4. The feature can be entirely optional, if any legal concerns are involved

PR #247, opened 2026-07-31
Map verified Debian license exceptions to SPDX

### LINKED ISSUE #246: Map verified Debian license exception synopses to SPDX
## Problem

License reporting introduced by [#153](https://github.com/siemens/debsbom/issues/153) and 
[#161](https://github.com/siemens/debsbom/pull/161) intentionally emits a package declaration 
only when every license synopsis in its machine-readable Debian copyright file can be mapped 
correctly to SPDX. This is a valid and [documented design decision](https://siemens.github.io/debsbom/design-decisions.html#license-information).

That conversion currently handles plain Debian license names, but not Debian synopses containing 
license exceptions. Raw forms such as `GPL-3+ with Bison exception` are not valid SPDX syntax 
and fail before the existing unknown-license lookup can translate them. As a result, packages 
with otherwise well-understood exceptions receive no declared license information.

This must retain the existing correctness guarantee: never emit a partial expression and never 
guess from a free-form exception name.

## Proposed mappings

Add exact, case-sensitive mappings for complete Debian license synopses whose license and 
exception texts have been verified:

| Debian synopsis | SPDX expression |
| --- | --- |
| `GPL-3+ with Bison exception` | `GPL-3.0-or-later WITH Bison-exception-2.2` |
| `GPL-3+ with Bison-2.2 exception` | `GPL-3.0-or-later WITH Bison-exception-2.2` |
| `GPL-2 with Linux-syscall-note exception` | `GPL-2.0-only WITH Linux-syscall-note` |
| `GPL-3+ with texinfo exception` | `GPL-3.0-or-later WITH Texinfo-exception` |
| `GPL-3.0-or-later-WITH-Autoconf-exception-macro` | `GPL-3.0-or-later WITH Autoconf-exception-macro` |
| `BSD-3-clause-Cambridge with BINARY LIBRARY-LIKE PACKAGES exception` | `BSD-3-Clause WITH PCRE2-exception` |

The source texts were checked in Debian 13/trixie copyright files within the `debian:13.6` 
base container for `tar`, `bash`, `findutils`, `systemd`, `diffutils`, `xz-utils`, and `pcre2`. 
The resulting exception identifiers are present in the SPDX License List and in `license-expression` 
30.4.1's SPDX symbol index.

The mapping must happen before parsing, because the Debian spelling is not necessarily valid 
SPDX `WITH` syntax. It should use a separate whole-synopsis table rather than fuzzy replacement 
of words inside expressions.

## Why exact matching is required

Debian exception names are free-form and are not stable identifiers. In particular, similarly named Autoconf forms are not equivalent:

- Nettle's `GPL-3+ with Autoconf exception` text matches SPDX `Autoconf-exception-generic`.
- Diffutils uses a similarly named form whose text differs and does not exactly match an SPDX exception.
- Findutils' `Autoconf-data exception` forms map to different SPDX generic exception identifiers depending on the GPL version.

Therefore broad Autoconf aliases, case folding, separator normalization, and partial/fuzzy 
exception matching would weaken the 100% correctness requirement and should not be added.

## Expected behavior

- A copyright file containing only plain licenses and supported exact exception synopses 
  emits the complete combined SPDX expression once the `debsbom generate` command is used 
  with `--with-licenses` flag.
- SPDX writes that expression to `licenseDeclared`.
- CycloneDX writes it to `licenses` with `acknowledgement` set to `declared`.
- Any unknown or ambiguous exception still rejects all license information for that source 
  package.
- A supported exception combined with unresolved `public-domain`, a custom license, an empty 
  synopsis, or another unknown term still rejects all license information for that source 
  package.

## Debian 13.6 impact

An amd64 `debian:13.6` rootfs has 78 installed binary packages represented by 120 source components 
once `Built-Using` and `Static-Built-Using` sources are included. Under the existing all-or-nothing 
policy:

- Before these aliases: 24/120 source components have declared license expressions (20.0%).
- After these aliases: 26/120 source components have declared license expressions (21.7%).
- Among the 55 sources corresponding directly to installed packages, coverage rises from 
  24/55 (43.6%) to 26/55 (47.3%).
- `gmp` and `tar` become newly complete.

The aliases also occur in `bash`, `diffutils`, `findutils`, `pam`, `pcre2`, `systemd`, and 
`xz-utils`, but those packages retain other unresolved terms. This is expected: no known 
conjunct should be silently discarded. Resolving `public-domain` declarations separately 
would make two additional packages (`pcre2` and `systemd`) benefit from these exception mappings.

## Scope

In scope:

- The six exact mappings above.
- Focused conversion and generated SPDX/CycloneDX tests.
- Documentation of exact exception handling and the all-or-nothing behavior.
- Strech goal: analyze all Debian 13 copyright files for additional exact exception synopses 
  and add them to the mapping table.

Out of scope:

- Partial license expressions.
- Fuzzy exception-name parsing or normalization.
- Ambiguous Autoconf/Automake name forms.
- Public-domain representation.
- Custom `LicenseRef-*` generation.
- License text scanning or equivalence inference at runtime.

## Duplicate search

I could not find any open issues, open pull requests, and prior license-related pull requests 
implementing verified exception aliases. Moreover [[license expression handling logic](https://github.com/hakandilek/debsbom/blob/main/src/debsbom/apt/copyright.py#L203)]
contains a `TODO` marker to address this exact problem:

```python
    @cached_property
    def _spdx_license_expressions(self) -> list[LicenseExpression]:
        """Return all licenses as SPDX license expressions."""
        licensing = get_spdx_licensing()
        exprs = []
        for lic in self.licenses():
            if not lic.synopsis:
                raise UnknownLicenseError("only license text is available")

            expr = self._convert_expression(lic.synopsis)
            spdx_lic = licensing.parse(expr)
            unknown_keys = licensing.unknown_license_keys(spdx_lic)
            # TODO: how do we handle `public-domain` licensing?
            # TODO: license exceptions
            if len(unknown_keys) > 0:
```

# Willing to contribute

Happy to open a PR for your review.

hakandilek:
## Summary

Add exact, case-sensitive mappings for six verified Debian license exception synopses:

| Debian synopsis | SPDX expression |
| --- | --- |
| `GPL-3+ with Bison exception` | `GPL-3.0-or-later WITH Bison-exception-2.2` |
| `GPL-3+ with Bison-2.2 exception` | `GPL-3.0-or-later WITH Bison-exception-2.2` |
| `GPL-2 with Linux-syscall-note exception` | `GPL-2.0-only WITH Linux-syscall-note` |
| `GPL-3+ with texinfo exception` | `GPL-3.0-or-later WITH Texinfo-exception` |
| `GPL-3.0-or-later-WITH-Autoconf-exception-macro` | `GPL-3.0-or-later WITH Autoconf-exception-macro` |
| `BSD-3-clause-Cambridge with BINARY LIBRARY-LIKE PACKAGES exception` | `BSD-3-Clause WITH PCRE2-exception` |

The aliases are applied before SPDX parsing because Debian exception names are free-form and may not use valid SPDX syntax. No fuzzy matching, normalization, or partial declarations are introduced.

Fixes #246 

This follows the correctness policy established in:

- https://github.com/siemens/debsbom/issues/153
- https://github.com/siemens/debsbom/pull/161

## Implementation

- Map only complete synopses whose license and exception texts were verified.
- Preserve package-level all-or-nothing rejection.
- Combine expressions through the licensing API for compatibility with the dependency pins.
- Document licese exception handling and remaining limitations within [design decisions](./docs/source/design-decisions.rst).
- Added relevant test cases to cover both SPDX/CycloneDX.

## Out of scope

Partial expressions, fuzzy exception parsing, ambiguous Autoconf/Automake forms, public-domain representation, and custom `LicenseRef-*` generation.

### DIFF
--- a/src/debsbom/apt/copyright.py
+++ b/src/debsbom/apt/copyright.py
@@ -13,6 +13,7 @@ from debian.copyright import (
 from license_expression import LicenseExpression, Licensing, get_spdx_licensing
 import logging
 from pathlib import Path
+import re
 
 logger = logging.getLogger(__name__)
 
@@ -126,6 +127,30 @@ WELL_KNOWN_EXPRESSIONS = {
     "Zope-2": "ZPL-2.0",
 }
 
+# Debian exception names are free-form. Only map complete license names with
+# exceptions whose license and exception texts have been verified.
+WELL_KNOWN_EXCEPTIONS = {
+    "GPL-3+ with Bison exception": "GPL-3.0-or-later WITH Bison-exception-2.2",
+    "GPL-3+ with Bison-2.2 exception": "GPL-3.0-or-later WITH Bison-exception-2.2",
+    "GPL-2 with Linux-syscall-note exception": "GPL-2.0-only WITH Linux-syscall-note",
+    "GPL-3+ with texinfo exception": "GPL-3.0-or-later WITH Texinfo-exception",
+    "GPL-3.0-or-later-WITH-Autoconf-exception-macro": (
+        "GPL-3.0-or-later WITH Autoconf-exception-macro"
+    ),
+    "BSD-3-clause-Cambridge with BINARY LIBRARY-LIKE PACKAGES exception": (
+        "BSD-3-Clause WITH PCRE2-exception"
+    ),
+    "GPL-3+-WITH-BISON-EXCEPTION": "GPL-3.0-or-later WITH Bison-exception-2.2",
+}
+
+_EXCEPTIONS = {name.lower(): expression for name, expression in WELL_KNOWN_EXCEPTIONS.items()}
+_EXCEPTIONS_RE = re.compile(
+    r"(?<![\w+.-])(?:{})(?![\w+.-])".format(
+        "|".join(re.escape(name) for name in sorted(WELL_KNOWN_EXCEPTIONS, key=len, reverse=True))
+    ),
+    re.IGNORECASE,
+)
+
 
 class UnknownLicenseError(Exception):
     """License is unknown to the SPDX standard."""
@@ -144,6 +169,8 @@ class Copyright(DebCopyright):
     @classmethod
     def _convert_expression(cls, line: str) -> str:
         """Convert a Debian license expression to the equivalent SPDX syntax."""
+
+        line = _EXCEPTIONS_RE.sub(lambda match: _EXCEPTIONS[match.group().lower()], line)
 
         # in the Debian copyright format ',' are used for disambiguation
         # that means (with 'and' having precedence over 'or'):
@@ -200,7 +227,6 @@ class Copyright(DebCopyright):
             spdx_lic = licensing.parse(expr)
             unknown_keys = licensing.unknown_license_keys(spdx_lic)
             # TODO: how do we handle `public-domain` licensing?
-            # TODO: license exceptions
             if len(unknown_keys) > 0:
                 # unknown keys in the license expression, try to replace them
                 # with well-known representations
--- a/src/debsbom/generate/cdx.py
+++ b/src/debsbom/generate/cdx.py
@@ -15,7 +15,7 @@ from cyclonedx.model import HashAlgorithm as cdx_hashalgo
 from cyclonedx.model import HashType as cdx_hashtype
 from cyclonedx.model.license import LicenseAcknowledgement, LicenseExpression, LicenseRepository
 from datetime import datetime
-from license_expression import ExpressionError
+from license_expression import AND, ExpressionError
 import logging
 from sortedcontainers import SortedSet
 from uuid import UUID, uuid4
@@ -109,10 +109,10 @@ def cdx_package_repr(
             )
         if package.copyright:
             try:
-                licenses = package.copyright.spdx_license_expressions()
-                combined = next(licenses)
-                for lic in licenses:
-                    combined &= lic
+                licenses = list(package.copyright.spdx_license_expressions())
+                combined = licenses[0]
+                if len(licenses) > 1:
+                    combined = AND(*licenses)
                 expression = LicenseExpression(
                     str(combined.simplify()), acknowledgement=LicenseAcknowledgement.DECLARED
                 )
--- a/src/debsbom/generate/spdx.py
+++ b/src/debsbom/generate/spdx.py
@@ -5,7 +5,7 @@
 from collections.abc import Callable, Iterable
 from datetime import datetime
 from importlib.metadata import version
-from license_expression import ExpressionError
+from license_expression import AND, ExpressionError
 import logging
 import spdx_tools.spdx.model.actor as spdx_actor
 import spdx_tools.spdx.model.document as spdx_document
@@ -185,10 +185,10 @@ def spdx_package_repr(package: Package, vendor: str = "debian") -> spdx_package.
         licenses_decl = SpdxNoAssertion()
         if package.copyright:
             try:
-                licenses = package.copyright.spdx_license_expressions()
-                combined = next(licenses)
-                for lic in licenses:
-                    combined &= lic
+                licenses = list(package.copyright.spdx_license_expressions())
+                combined = licenses[0]
+                if len(licenses) > 1:
+                    combined = AND(*licenses)
                 licenses_decl = combined.simplify()
             except (ExpressionError, UnknownLicenseError) as e:
                 logger.debug(f"no SPDX license expression for {package}: {e}")
--- a/docs/source/design-decisions.rst
+++ b/docs/source/design-decisions.rst
@@ -217,11 +217,13 @@ License Information
 
 As the specifiers in Debian copyright files are not SPDX identifiers a conversion table based on the `syntax specification <https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-specification>`__ is used.
 
+Debian license exception names are free-form, so complete license names containing exceptions are mapped only when both the license and exception have been verified. These mappings are case-insensitive, as required by the Debian copyright format, but ``debsbom`` does not normalize separators or partially or fuzzily match exception names.
+
 For SPDX SBOMs the information is placed in the `declared licenses field <https://spdx.github.io/spdx-spec/v2.3/package-information/#715-declared-license-field>`__; and for CDX SBOMs in the `licenses field <https://cyclonedx.org/docs/1.6/json/#components_items_licenses>`__ with the ``acknowledgement`` set to ``declared``. This is done this way because ``debsbom`` does not perform any analysis of included licenses and just trusts what is stated by the package authors. This can be expressed in the SBOMs with the above approach. For CDX it could alternatively be placed in the `evidence.licenses field <https://cyclonedx.org/docs/1.6/json/#components_items_evidence_licenses>`__ as there is, as far as we can tell, no clear distinction between the meaning of the two if all ``acknowledgments`` are set to ``declared``. We decided to put it directly in ``licenses``, as that is the first place a user would expect it.
 
 Limitations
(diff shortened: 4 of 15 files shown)


src/debsbom/apt/copyright.py
```diff
@@ -196,11 +211,12 @@ def _spdx_license_expressions(self) -> list[LicenseExpression]:
             if not lic.synopsis:
                 raise UnknownLicenseError("only license text is available")
 
-            expr = self._convert_expression(lic.synopsis)
+            expr = WELL_KNOWN_EXCEPTIONS.get(lic.synopsis)
```
Urist-McGit: `expr` contains the whole line of the license synopsis. So this works only if the exception is the only license in the synopsis and would break if the expression would be e.g. `GPL-3+ with Bison exception or Expat`.

src/debsbom/generate/cdx.py
```diff
@@ -112,7 +112,7 @@ def cdx_package_repr(
                 licenses = package.copyright.spdx_license_expressions()
                 combined = next(licenses)
                 for lic in licenses:
-                    combined &= lic
+                    combined = get_spdx_licensing().AND(combined, lic)
```
Urist-McGit: This does not seem related to the license expression with exceptions. Is there a reason why this was added?

src/debsbom/generate/spdx.py
```diff
@@ -188,7 +188,7 @@ def spdx_package_repr(package: Package, vendor: str = "debian") -> spdx_package.
                 licenses = package.copyright.spdx_license_expressions()
                 combined = next(licenses)
                 for lic in licenses:
-                    combined &= lic
+                    combined = get_spdx_licensing().AND(combined, lic)
```
Urist-McGit: Same as above: why was this changed?

src/debsbom/apt/copyright.py
hakandilek: Ah, you're right. `WELL_KNOWN_EXCEPTIONS.get(lic.synopsis)` matches the entire License: line, so it only fires when the exception is the whole synopsis.

Replacing whole license names anywhere in the synopsis, before parsing, instead of matching the whole line would fix this.

Some Regex-magic would do the job like:

```python
_EXCEPTIONS = {name.lower(): expr for name, expr in WELL_KNOWN_EXCEPTIONS.items()}

# only replace complete license names, never a part of a longer name
_EXCEPTIONS_RE = re.compile(
    r"(?<![\w+.-])(?:"
    + "|".join(re.escape(name) for name in sorted(_EXCEPTIONS, key=len, reverse=True))
    + r")(?![\w+.-])",
    re.IGNORECASE,
)

# and have the substitution as the first step of the existing conversion:
line = _EXCEPTIONS_RE.sub(lambda m: _EXCEPTIONS[m.group().lower()], line)
```

Subtly complex but should work. I'll amend a fix.

src/debsbom/generate/cdx.py
hakandilek: Ah I see, this change is related but do to a different problem. I should have splitted this into a second commit or even a second PR.

This is a separate bug due to a typing error: `combined` and `lic` aren't strings but objects of **different** types: `LicenseSymbol` and `LicenseWithExceptionSymbol`. `LicenseSymbol` implements `AND` but `LicenseWithExceptionSymbol` does not.

So if we've the license sequence like this, it'll work:

```text
License: GPL-3+                       <- accumulator starts here, LicenseSymbol
License: GPL-3+ with Bison exception  <- LicenseSymbol & LicenseWithExceptionSymbol,
                                         fine, result is an AND node
```

but if the exception first comes up in the compyright file, it'll fail:

```text
License: GPL-3+ with Bison exception  <- accumulator starts here, LicenseWithExceptionSymbol
License: Expat                        <- LicenseWithExceptionSymbol & LicenseSymbol 
                                         -> TypeError
```

So the fix is the licensing object to build the node rather than the left operand.

I'll split this into a second commit. Please let me know what you think.

Urist-McGit: This looks more like a bug in `license_expression` to me, maybe we should report it. Since we previously didnt have expressions with exceptions it can stay in the same commit.

hakandilek: > maybe we should report it.

Good point, I'll do it!

> Since we previously didnt have expressions with exceptions it can stay in the same commit.

Hah, just pushed it in a separate commit explaining the details. The commit 765cff0cc4e6d6729a9f35608989fe64e3b2dac2 also contains a much defensive expression handling and some test cases

Urist-McGit: Either way is fine ;)

src/debsbom/apt/copyright.py
hakandilek: Please have a look at the changes on e3153006a116735b861d4b6f626376ebafb5cc7b

src/debsbom/apt/copyright.py
```diff
@@ -145,6 +170,8 @@ def __init__(self, path: Path):
     def _convert_expression(cls, line: str) -> str:
         """Convert a Debian license expression to the equivalent SPDX syntax."""
 
+        line = _EXCEPTIONS_RE.sub(lambda match: _EXCEPTIONS[match.group().lower()], line)
```
Urist-McGit, line 173: This is a very expensive operation. We have 7 exceptions right now, and I can see a ~30% time increase from `main` right now. I think this will increase linearly with the amount of exceptions that we consider.

```
$ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
$ time $(which debsbom) generate -t spdx --with-licenses

real	1m37.183s
user	1m36.678s
sys	0m0.477s
$ git checkout -
Switched to branch 'feat/license-exceptions'
$ time $(which debsbom) generate -t spdx --with-licenses

real	2m2.403s
user	2m1.755s
sys	0m0.613s
```

Maybe it would be a better idea to try these replacements on the unknown symbols? That would decrease the candidates and the hay for the regexp significantly. We probably would need to do some work reconstructing the expressions as I think the unknown license keys are split by whitespace, but it should be possible.

Urist-McGit: > I think this will increase linearly with the amount of exceptions that we consider.

Just tested a bit more, it seems like the timeloss is not dependent on the number of exceptions. Now I wonder just how that regex works.

Urist-McGit: Ok, further research was done: the time loss goes back solely to the `licensing.AND` change. So I guess this is fine. Goes to show that consistent benchmarking is important and how hard it is to spot performance regressions.

hakandilek: >  the time loss goes back solely to the `licensing.AND` change.

Ah I see. This is actually calling `get_spdx_licensing()` once per `Copyright` object multiple times. 

Shall we apply some caching?

Urist-McGit: It is indeed the call to `get_spdx_licensing` that is causing this regression. We should cache it then.

Urist-McGit: Can you merge dbba3bc09e3a229abdf41905a0a006019745e7aa into the first commit 765cff0cc4e6d6729a9f35608989fe64e3b2dac2? The performance regression was introduced in this commit, and the changes fit just fine in there.

Otherwise this looks good to merge, thank you!

hakandilek: > Can you merge https://github.com/siemens/debsbom/commit/dbba3bc09e3a229abdf41905a0a006019745e7aa into the first commit https://github.com/siemens/debsbom/commit/765cff0cc4e6d6729a9f35608989fe64e3b2dac2?

Done.

Assigning it to you to merge. Thanks for the patience! 🙂

Urist-McGit (approved):

PR #249, opened 2026-07-31
Parse copyright files atomically before yielding licenses

### LINKED ISSUE #248: Partially parsed copyright files emit an incomplete declared license
## Summary

`Copyright.licenses()` is a generator that yields licenses *while* parsing. If
python-debian raises after some licenses have already been yielded, the error is
caught but the values have already escaped. The caller therefore receives a
truncated list that is indistinguishable from a successful parse, and debsbom
emits a **declared license that is narrower than what the package actually
declares**.

This is a fail-open defect: it does not merely lose data, it asserts something
untrue. That conflicts with the "100% correct" requirement established in #153.

## Root cause

`src/debsbom/apt/copyright.py:199-207`

```python
def licenses(self) -> Iterable[License]:
    with open(self._path) as f:
        try:
            yield from self._parse_licenses(DebCopyright(f))   # values escape before the raise
        except NotMachineReadableError:
            logger.debug(...)
        except (MachineReadableFormatError, ValueError):
            logger.debug(...)
```

`_parse_licenses` iterates `all_files_paragraphs()` lazily and accesses
`paragraph.license` per paragraph. A malformed paragraph raises
`MachineReadableFormatError` at that point — after earlier paragraphs have
already been yielded. `except` then swallows it, so `_spdx_license_expressions`
combines only the prefix and reports success.

## Evidence

Two source packages in trixie emit a strictly narrower expression than declared:

- `libswingx-java` line 29 — TAB before the Apache licence URL, in the **2nd of 4**
  paragraphs, so only the 1st licence survives.
- `jclic` line 52 — TAB inside the **7th and last** paragraph, so 6 of 7 survive.

| Package | `Files` paragraphs | Emitted | Silently dropped |
| --- | ---: | --- | --- |
| `libswingx-java=1:1.6.2-5` | 4 | `LGPL-2.1-or-later` | `Apache-2.0`, `LPGL-2.1+ or BSD-3-clause`, `GPL-2+` |
| `jclic=0.3.2.10-1.1` | 7 | `GPL-2.0-or-later AND LGPL-2.1-or-later` | `BSD-4` |

In both cases the trigger is a continuation line indented with a TAB instead of
a space, which python-debian rejects with
`MachineReadableFormatError: continued line must begin with " "`.

## Minimal reproducer

`copyright` (the line before `http://www.apache.org/licenses/LICENSE-2.0` begins with a literal **TAB**):

```
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
Upstream-Name: late-parse-error

Files: first/*
Copyright: 2026 Siemens
License: Expat
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files.

Files: second/*
Copyright: 2026 Siemens
License: Apache-2.0
    Licensed under the Apache License, Version 2.0 (the "License");
    you may obtain a copy of the License at
 .
	http://www.apache.org/licenses/LICENSE-2.0
 .
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS.

Files: third/*
Copyright: 2026 Siemens
License: GPL-2+
 Debian packaging is licensed under the GPL version 2 or later.
```

```python
from pathlib import Path
from debsbom.apt.copyright import Copyright

cr = Copyright(Path("copyright"))
print([lic.synopsis for lic in cr.licenses()])
print([str(e) for e in cr.spdx_license_expressions()])
```

**Actual**

```
['Expat']
['MIT']
```

**Expected** — the file declares `Expat`, `Apache-2.0` and `GPL-2+`; emitting only
`MIT` claims a narrower licence than declared.

Replacing that single TAB with a space makes the same file parse completely and
yield all three licences, which isolates the trigger.

## Suggested fix

Parse atomically, then yield. Consumption stays lazy; only the parse becomes
eager, which costs nothing because `_spdx_license_expressions` is a
`cached_property` that drains the iterator anyway.

```python
def licenses(self) -> Iterable[License]:
    """Return all licenses found in the copyright file."""
    with open(self._path) as f:
        try:
            licenses = list(self._parse_licenses(DebCopyright(f)))
        except NotMachineReadableError:
            logger.debug(f"non-machine-readable copyright file: {self._path}")
            return
        except (MachineReadableFormatError, ValueError):
            logger.debug(f"bad format for machine-readable copyright file: {self._path}")
            return
    yield from licenses
```

hakandilek:
Fixes #248.

## Problem

`Copyright.licenses()` yielded licenses *while* python-debian was still parsing.
When parsing raised after one or more licenses had already been yielded, the
error was caught but the values had already escaped. The caller received a
truncated list indistinguishable from a complete parse, so debsbom emitted a
declared license **narrower than the package actually declares**.

This behaviour does not merely lose information, it asserts something
untrue, conflicting with the all-or-nothing guarantee from #153.

The malformed test case declares `Expat`, `Apache-2.0` and `GPL-2+`:

| | `licenses()` | `spdx_license_expressions()` |
| --- | --- | --- |
| before | `['Expat']` | `['MIT']` |
| after | `[]` | raises `UnknownLicenseError` |

This is triggered by a TAB-indented continuation line that python-debian rejects
with `continued line must begin with " "`.

## Fix

The parse is called inside the existing `try`, and yield only after it
completes.

## Commits

1. `9edfcc5` contains the test fixture that fail on the defect.
2. `a6c1bfb` the fix with a positive-control fixture and test.

The positive control matters: both tests in the first commit assert *absence*,
so a fix that simply returned nothing would pass them.

### DIFF
--- a/src/debsbom/apt/copyright.py
+++ b/src/debsbom/apt/copyright.py
@@ -200,11 +200,14 @@ class Copyright(DebCopyright):
         """Return all licenses found in the copyright file."""
         with open(self._path) as f:
             try:
-                yield from self._parse_licenses(DebCopyright(f))
+                licenses = list(self._parse_licenses(DebCopyright(f)))
             except NotMachineReadableError:
                 logger.debug(f"non-machine-readable copyright file: {self._path}")
+                return
             except (MachineReadableFormatError, ValueError):
                 logger.debug(f"bad format for machine-readable copyright file: {self._path}")
+                return
+        yield from licenses
 
     def _parse_licenses(self, copyright: DebCopyright) -> Iterable[License]:
         lic = copyright.header.license
--- a/tests/data/late-parse-error-copyright
+++ b/tests/data/late-parse-error-copyright
@@ -1,0 +1,24 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: late-parse-error
+
+Files: first/*
+Copyright: 2026 Siemens
+License: Expat
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files.
+
+Files: second/*
+Copyright: 2026 Siemens
+License: Apache-2.0
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may obtain a copy of the License at
+ .
+	http://www.apache.org/licenses/LICENSE-2.0
+ .
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS.
+
+Files: third/*
+Copyright: 2026 Siemens
+License: GPL-2+
+ Debian packaging is licensed under the GPL version 2 or later.
--- a/tests/data/late-parse-ok-copyright
+++ b/tests/data/late-parse-ok-copyright
@@ -1,0 +1,24 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: late-parse-ok
+
+Files: first/*
+Copyright: 2026 Siemens
+License: Expat
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files.
+
+Files: second/*
+Copyright: 2026 Siemens
+License: Apache-2.0
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may obtain a copy of the License at
+ .
+ http://www.apache.org/licenses/LICENSE-2.0
+ .
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS.
+
+Files: third/*
+Copyright: 2026 Siemens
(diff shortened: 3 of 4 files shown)


Urist-McGit:
Could you add a fixes tag for a6c1bfbadd3ed5fa634d7e11ba41239ef07faa19? The introducing commit should be 0ab9f9e797a25196cfd71115f11861cd288c5363

Urist-McGit:
And please reorganize the commits so the failing test comes after the bug fix. We want to have our tests always run without failure.

hakandilek:
> please reorganize the commits so the failing test comes after the bug fix. We want to have our tests always run without failure.

Good to know that. Done.

> Could you add a fixes tag for https://github.com/siemens/debsbom/commit/a6c1bfbadd3ed5fa634d7e11ba41239ef07faa19? The introducing commit should be https://github.com/siemens/debsbom/commit/0ab9f9e797a25196cfd71115f11861cd288c5363

.. and done.

Urist-McGit (approved):

PR #253, opened 2026-08-03
Fall back to sibling binaries for source copyright lookup

hakandilek:
## Problem

Commit [`49210a1`](https://github.com/siemens/debsbom/commit/49210a1) added source-level de-duplication so copyright data would be scanned once per source package. Before this change, the source was recorded as processed before the selected binary's copyright path was checked:

```python
try:
    src_processed.add(src_hash)
    src_pkg.copyright = cr_dir.copyright(bin_pkg)
except FileNotFoundError:
    logger.debug(f"no copyright information for {bin_pkg}")
    continue
```

If `/usr/share/doc/<package>/copyright` does not resolve for that binary, the lookup raises `FileNotFoundError`, but later binaries from the same source are skipped. This changed the intended "scan copyright data once per source" behavior into "attempt one binary per source."

`filter_binaries()` preserves dictionary insertion order. Source records can be injected before this loop, but they are filtered out, while binary records retain dpkg status order. The selected binary is therefore deterministic, but its position is unrelated to whether its copyright path resolves.

## Debian validation

This is not evidence of a current Debian packaging bug. [Debian Policy section 12.5](https://www.debian.org/doc/debian-policy/ch-docs.html#copyright-information) requires every binary package to provide a resolvable copyright path and permits its documentation directory to link to a same-source sibling on which it has a strict dependency. debsbom already handles the normal case because `Path.is_file()` follows symbolic links.

No archive-produced example was found in the published Trixie, Trixie updates, or Trixie backports `Packages` and `Contents` indices. Across 72,395 checked `amd64` and `all` binary records, every package either ships its own copyright file or provides a documentation-directory link. The downloaded indices were verified against their signed `InRelease` metadata. The standard `debian:13-slim` documentation filter also preserves these links.

The defect is relevant to transformed root filesystems that no longer exactly match their packages. For example, documentation pruning or independent extraction can leave the first binary's link dangling while another same-source binary's copyright file remains. The current control flow then discards available source-level license data based only on package order.

## Fix

Record a source as processed only after a copyright path resolves:

```python
try:
    src_pkg.copyright = cr_dir.copyright(bin_pkg)
    src_processed.add(src_hash)
except FileNotFoundError:
    logger.debug(f"no copyright information for {bin_pkg}")
    continue
```

This preserves the optimization: only the first successful copyright lookup is retained for each source. The only additional work is checking remaining sibling paths after a miss.

## Tests

- `copyright-sibling` models a dangling `--link-doc` path in a transformed rootfs while another sibling still ships the copyright file. Without the fix, the source gets `NOASSERTION`; with the fix, it gets `MIT`.
- `copyright-linked-doc` covers the normal Policy-compliant case where the documentation link resolves. It continues to produce `MIT`.

Both tests pass with the fix.

## Commits

1. Fix the lookup so every commit remains green.
2. Add the dangling-link regression and resolvable-link coverage.


hakandilek:
@Urist-McGit I'm not really sure about this one yet. I just need to make a through analysis to figure out if there are valid examples of that. I'll create an issue once I've some data.

Urist-McGit:
If a binary package does not distribute its copyright files this seems like a packaging bug. Does this actually occur?

hakandilek:
> If a binary package does not distribute its copyright files this seems like a packaging bug. Does this actually occur?

You're right and this doesn't occur in Debian, at least I could not identify it in Trixie, so I've dropped that overall claim and adapted the PR description to the actual latent defect. 

I'm not sure if it should be addressed as bug, if then I'd file an issue as well.

Urist-McGit:
@hakandilek have you had a chance to look more into this?

hakandilek:
@Urist-McGit I've been on vacation the last two weeks. I'll get back on it mid-next week once I'm back.

Urist-McGit (approved):

PR #254, opened 2026-08-06
Add `distro` qualifiers to PURLs

Urist-McGit:
Add the distro qualifier to PURLs if it is available. It contains the codename of the distribution as specified in the corresponding Release file in the apt cache. Also consider the distro for security scanning.

As this qualifier is not specified but in some examples, we might want to wait before merging this. [1]

[1] https://github.com/package-url/purl-spec/blob/main/types/deb-definition.json

### DIFF
--- a/src/debsbom/commands/security_scan.py
+++ b/src/debsbom/commands/security_scan.py
@@ -135,7 +135,9 @@ class SecurityScanCmd(SbomInput, PkgStreamInput):
             )
         )
         parser.add_argument(
-            "--distro", default="trixie", help="Debian distribution to check (default: %(default)s)"
+            "--distro",
+            default="trixie",
+            help="Debian distribution to check if package has no distribution information (default: %(default)s)",
         )
         parser.add_argument(
             "--update-db",
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -483,12 +483,12 @@ class SourcePackage(Package):
         self.copyright = copyright
 
     def __hash__(self):
-        return hash(self.purl())
+        return hash((self.name, self.version))
 
     def __eq__(self, other):
         # For compatibility reasons
         if other.is_source():
-            return self.purl() == other.purl()
+            return (self.name, self.version) == (other.name, other.version)
         return NotImplemented
 
     def purl(self, vendor="debian") -> PackageURL:
@@ -635,11 +635,15 @@ class BinaryPackage(Package):
         self.status = status
 
     def __hash__(self):
-        return hash(self.purl())
+        return hash((self.name, self.version, self.architecture))
 
     def __eq__(self, other):
         if other.is_binary():
-            return self.purl() == other.purl()
+            return (self.name, self.version, self.architecture) == (
+                other.name,
+                other.version,
+                other.architecture,
+            )
         return NotImplemented
 
     def purl(self, vendor="debian") -> PackageURL:
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -330,13 +330,17 @@ class Package(ABC):
         if not purl.type == "deb":
             raise RuntimeError("Not a debian purl", purl)
         if purl.qualifiers.get("arch") == "source":
-            return SourcePackage(purl.name, purl.version)
+            package = SourcePackage(purl.name, purl.version)
+            package.distro = purl.qualifiers.get("distro")
+            return package
         else:
-            return BinaryPackage(
+            package = BinaryPackage(
                 name=purl.name,
                 architecture=purl.qualifiers.get("arch"),
                 version=purl.version,
             )
+            package.distro = purl.qualifiers.get("distro")
+            return package
 
     @classmethod
     def inject_src_packages(cls, binpkgs: Iterable["BinaryPackage"]) -> Iterable["Package"]:
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -496,10 +496,16 @@ class SourcePackage(Package):
 
     def purl(self, vendor="debian") -> PackageURL:
         """Return the PURL of the package."""
-        purl = "pkg:deb/{}/{}@{}?arch=source".format(vendor, self.name, self.version)
+        qualifiers = {"arch": "source"}
         if self.distro:
-            purl = purl + "&distro={}".format(self.distro)
-        return PackageURL.from_string(purl)
+            qualifiers["distro"] = self.distro
+        return PackageURL(
+            type="deb",
+            namespace=vendor,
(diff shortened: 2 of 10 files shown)


src/debsbom/dpkg/package.py
```diff
@@ -493,9 +496,10 @@ def __eq__(self, other):
 
     def purl(self, vendor="debian") -> PackageURL:
         """Return the PURL of the package."""
-        return PackageURL.from_string(
-            "pkg:deb/{}/{}@{}?arch=source".format(vendor, self.name, self.version)
-        )
+        purl = "pkg:deb/{}/{}@{}?arch=source".format(vendor, self.name, self.version)
+        if self.distro:
+            purl = purl + "&distro={}".format(self.distro)
+        return PackageURL.from_string(purl)
```
fmoessbauer: Didn't the PackageURL library has support for building a purl from its components, instead of creating a purl-string and parsing it?

Urist-McGit: It has and it can be done easily, I will add another commit with the PURL construction from components.

hakandilek:
As you've also noticed, unfortunately there's no aligned `distro` qualifier within purl.

Different binary SBOM analysis tools also deliver different type of distro values:

- wiz generates `debian-<major>.<minor>`  - e.g., `distro=debian-12.12`
- mend has `debian-<major>` - e.g., `distro=debian-12`
- syft also `debian-<major>` - e.g., `distro=debian-13`

The distro information is helpful in a vulnerability management perspective to match the packages to the corresponding entry in the security tracker. I think `<major>` would be enough and `<major>.<minor>` detail is not really needed.

fmoessbauer:
> As you've also noticed, unfortunately there's no aligned `distro` qualifier within purl.

The goal of this MR is to provide a PoC of what technically can be "scanned" just based on the apt data. Note, that the debsbom uses-cases are more complex than having a single rootfs from a single distribution. Reading the distro from `/etc/os-release` (as others are doing) is not sufficient.

The PoC should help to come up with a standardized version of the `distro` qualifier - or at least help to get the discussion started.

> The distro information is helpful in a vulnerability management perspective to match the packages to the corresponding entry in the security tracker. 

Yes, exactly. That's why we want to have it - in a standardized way.

Urist-McGit:
To add on to this: the security tracker JSON database has the releases identified by their codenames, as it is done in this PR.

src/debsbom/securityscan/scanner.py
```diff
@@ -72,7 +72,7 @@ def candidates(self, p: SourcePackage) -> Iterator[CveEntry]:
         if not vulns:
             return
         for k, v in vulns.items():
-            v_distr = v["releases"].get(self.distro)
+            v_distr = v["releases"].get(p.distro if p.distro else self.distro)
```
fmoessbauer: isn't that simply `p.distro or self.distro`?

src/debsbom/commands/security_scan.py
```diff
@@ -135,7 +135,9 @@ def setup_parser(cls, parser):
             )
         )
         parser.add_argument(
-            "--distro", default="trixie", help="Debian distribution to check (default: %(default)s)"
+            "--distro",
+            default="trixie",
+            help="Debian distribution to check if package has no distribution information (default: %(default)s)",
```
woefe, line 140: Is trixie a safe fallback here? Does it make sense to first attempt do detect the distro in another way e.g. /etc/*-release files?

fmoessbauer: As running `debsbom sec-scan` usually happens on a different system, we have no access to the original `/etc/os-release` anymore. This data would have to be captured in the SBOM, but there is no standard way to do this.

fmoessbauer (approved):

fmoessbauer (changes requested):
We also need to propagate the distro argument of binary packages to ones created by `static-built-using` or `built-using`. These always (usually in case of isar) come from the same distribution, but if we don't have the packages in the apt cache, they are currently not annotated with `distro=...`.

These annotations should be weak, so if we find the package in the apt-cache later on, we can use the definitive one.

src/debsbom/commands/security_scan.py
woefe: Would it make sense to set a distro version fallback on the `generate` side then? I see `generate` has `--distro-*` parameters which set fields of `metadata.component` of the resulting CycloneDX, but there is also `--base-distro-vendor` that can change the generated PURLs to `ubuntu`. Maybe a `--base-distro-version` parameter that sets a fallback for the `distro` qualifier, if the apt chache files aren't available?

fmoessbauer:
> We also need to propagate the distro argument of binary packages to ones created by `static-built-using` or `built-using`. 

This is now implemented and pushed on top. @Urist-McGit: please review carefully and fold in the commit that simplifies package construction.

src/debsbom/commands/security_scan.py
Urist-McGit: I think having the disitrbution as a metadata field is problematic. This would be another non-standard field that we use for consuming SBOMs, which we generally want to avoid. Additionally the distribution is not something that can be set on a per system level, it is attached to each specific package. A generator that knows the general distribution should just annotate any packages with the correct distribution.

fmoessbauer (approved):

Urist-McGit:
I rebased and squashed one of the commits. Please check again I did not make any mistakes, then we can merge.

src/debsbom/dpkg/package.py
```diff
@@ -651,6 +655,8 @@ def purl(self, vendor="debian") -> PackageURL:
         purl = "pkg:deb/{}/{}@{}".format(vendor, self.name, self.version)
         if self.architecture:
             purl = purl + "?arch={}".format(self.architecture)
+        if self.distro:
```
fmoessbauer, line 673: I just noticed, that this will break if we don't have an arch. Then the PURL would be `pkg:deb/debian/name@version&distro=foo`.

Urist-McGit: Ah true, but I think we can construct the PURL from its components instead. Will fix it

Urist-McGit: Wait, we already are constructing it correctly. I think the PURL constructors should handle the qualifiers properly. Did you actually produce the issue?

fmoessbauer: I just checked the commits: In 4ebf251edddcdd022df2f8b6aaa08239bec759c8 we fix this again. We better first do 4ebf251edddcdd022df2f8b6aaa08239bec759c8 and then add the distro qualifier.

Urist-McGit: I reordered the commits, now we always construct the PURL correctly.

fmoessbauer (approved):

PR #256, opened 2026-08-10
Derive the distro architecture from the dpkg status file

### LINKED ISSUE #255: `generate` fails on any rootfs without `/var/lib/dpkg/arch-native` (dpkg < 1.22.16)
`debsbom generate` aborts with exit 254 whenever `<root>/var/lib/dpkg/arch-native` is absent on distros with dpkg < 1.22.16:

```
ERROR:debsbom.cli:debsbom: error: Unable to auto detect distro architecture.
                          Set --distro-arch to dpkg architecture (e.g. amd64)
```

`<root>/var/lib/dpkg/arch-native` file is currently the **only** auto-detection strategy ([`_import_packages()` in `generate.py`](https://github.com/siemens/debsbom/blob/main/src/debsbom/generate/generate.py#L121)) and is only available since **dpkg 1.22.16**.

| image | dpkg | `arch-native` |
|---|---|---|
| `debian:11` | 1.20.14 | missing |
| `debian:12` | 1.21.23 | missing |
| `debian:13` | 1.22.21 | present |
| `ubuntu:20.04` | 1.19.7 | missing |
| `ubuntu:22.04` | 1.21.1 | missing |
| `ubuntu:24.04` | 1.22.6 | missing |
| `ubuntu:25.10` | 1.22.21 | present |

## Some findings and considerations

- Running natively on a Debian 12 or Ubuntu 24.04 host with the default `--root /` fails.
- The [documented](https://github.com/siemens/debsbom/blob/main/docs/source/examples.rst?plain=1#L31) [examples](https://github.com/siemens/debsbom/blob/main/docs/source/examples.rst?plain=1#L40) in [`docs/source/examples.rst`](https://github.com/siemens/debsbom/blob/main/docs/source/examples.rst?plain=1#L31) are also based on `bookworm` and would fail today.
- `--distro-arch` parameter is introduced as a workaround if the auto detection fails. The CI Pipeline in `.github/actions/pytest/action.yml` uses this parameter on the `ubuntu-24.04` runner (has no `arch-native`)

## Reproduction

```sh
mkdir -p /tmp/r12/var/lib/dpkg
docker run --rm --platform linux/amd64 --entrypoint sh debian:12 \
  -c 'cat /var/lib/dpkg/status' > /tmp/r12/var/lib/dpkg/status

debsbom generate --root /tmp/r12 -t cdx -o /tmp/sbom                     # exit 254
debsbom generate --root /tmp/r12 -t cdx -o /tmp/sbom --distro-arch amd64 # ok with the `--distro-arch`
```

## Suggestion

We can extend the `arch-native` detection to the status files (which debsbom
already parses) of good known packages. 

Resolution order:

1. `--distro-arch` (explicit)
2. `<root>/var/lib/dpkg/arch-native` -> unchanged, stays authoritative
3. `Architecture:` of the `dpkg` entry, then `base-files`
4. the single distinct concrete `Architecture` in the status file, if there is only one
5. otherwise `DistroArchUnknownError`, as today

The dpkg postinst writes `$DPKG_MAINTSCRIPT_ARCH` into
`arch-native`, and dpkg sets that variable to the package's own architecture
 so `arch-native` and the `Architecture:` of the
`dpkg` entry are **the same value**. Both marker packages are `Essential: yes` and
`Multi-Arch: foreign`, so they can never be co-installed for a foreign architecture
(`apt-get install dpkg:i386` → `dpkg : Conflicts: dpkg:i386`), which keeps cross-architecture
rootfs correct.

What do you think?

hakandilek:
## Summary

- derive the native architecture from the installed `dpkg` package when `arch-native` is unavailable
- preserve explicit `--distro-arch` and `arch-native` precedence
- keep erroring out when the `dpkg` package cannot be found
- leave `--from-pkglist` behavior unchanged
- remove the CI architecture override
- update rootfs examples to Debian Trixie
- document architecture resolution

## Motivation

`/var/lib/dpkg/arch-native` was introduced in dpkg 1.22.16, so Debian 12 and older Ubuntu releases cannot auto-detect the distribution architecture even though it is recorded in `/var/lib/dpkg/status`.

The `dpkg` post-installation script writes `$DPKG_MAINTSCRIPT_ARCH` to `arch-native`, which is the architecture of the installed `dpkg` package. Reading that package's `Architecture` field therefore relies on exactly the same fact rather than on a new heuristic.

Closes #255

### DIFF
--- a/.github/actions/pytest/action.yml
+++ b/.github/actions/pytest/action.yml
@@ -25,8 +25,7 @@ runs:
     env:
       COVERAGE_FILE: .coverage.smoke-generate.${{ inputs.artifact-identifier }}
     run: |
-      DISTRO_ARCH="$(dpkg --print-architecture)"
-      coverage run $(which debsbom) -v generate --distro-arch $DISTRO_ARCH -t spdx -t cdx -o sbom --validate
+      coverage run $(which debsbom) -v generate -t spdx -t cdx -o sbom --validate
 
   - name: smoke test merge
     shell: bash
--- a/src/debsbom/generate/generate.py
+++ b/src/debsbom/generate/generate.py
@@ -41,6 +41,12 @@ class DistroArchUnknownError(RuntimeError):
 
 
 class Debsbom:
+    # Package whose Architecture states the native architecture of the rootfs.
+    # The dpkg postinst writes $DPKG_MAINTSCRIPT_ARCH, which is the architecture
+    # of the installed dpkg package, to /var/lib/dpkg/arch-native. Both sources
+    # therefore report the same value.
+    _NATIVE_ARCH_MARKER = "dpkg"
+
     def __init__(
         self,
         distro_name: str,
@@ -110,6 +116,7 @@ class Debsbom:
         return metadata
 
     def _import_packages(self, stream=None):
+        from_rootfs = not stream
         if stream:
             packages_it = Package.parse_pkglist_stream(stream)
             # if we use packages from a stream we skip extended states since
@@ -119,13 +126,15 @@ class Debsbom:
             packages_it = Package.parse_status_file(self.root / "var/lib/dpkg/status")
             merge_ext_states = True
         if not self.distro_arch:
-            if self.root:
-                self.distro_arch = self._parse_distro_arch(self.root / "var/lib/dpkg/arch-native")
-            if not self.distro_arch:
-                raise DistroArchUnknownError()
-        logger.debug(f"distro arch is '{self.distro_arch}'")
+            self.distro_arch = self._parse_distro_arch(self.root / "var/lib/dpkg/arch-native")
 
         pkgdict = dict(map(lambda p: (hash(p), p), packages_it))
+        if not self.distro_arch and from_rootfs:
+            self.distro_arch = self._derive_distro_arch(pkgdict.values())
+        if not self.distro_arch:
+            raise DistroArchUnknownError()
+        logger.debug(f"distro arch is '{self.distro_arch}'")
+
         self.packages = self._merge_apt_data(
             pkgdict,
             inject_sources=packages_it.kind != PkgListType.STATUS_FILE,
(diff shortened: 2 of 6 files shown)


src/debsbom/generate/generate.py
```diff
@@ -41,6 +41,8 @@ def __init__(self):
 
 
 class Debsbom:
+    _NATIVE_ARCH_MARKER = "dpkg"
```
Urist-McGit, line 48: Please add a comment that explains what this exactly this.

.github/actions/pytest/action.yml
```diff
@@ -118,4 +117,3 @@ runs:
       path: .coverage.*
       include-hidden-files: true
       if-no-files-found: error
-
```
fmoessbauer, line 121: Unrelated change.

src/debsbom/generate/generate.py
```diff
@@ -141,6 +146,19 @@ def _parse_distro_arch(cls, arch_native_file: Path) -> str | None:
...
+        architectures = set()
+        for package in filter_binaries(packages):
+            architecture = package.architecture
+            if not architecture or architecture == "all":
+                continue
+            if package.name == cls._NATIVE_ARCH_MARKER:
+                return architecture
+            architectures.add(architecture)
+
+        return architectures.pop() if len(architectures) == 1 else None
```
Urist-McGit: I am not a fan of more heuristics honestly. Since the arch-native file is deployed by `dpkg` anyways it is clear that we rely on the same fact. I think it is best we only rely on `dpkg` on it as it should be always be available in a functioning Debian distribution.

Urist-McGit:
Commit 43328fb67d757cb4e23fc85b3e524f0967e40fc5 is not really a fix. As I see it you just extend the functionality to not rely solely on the arch-native file.

Urist-McGit (changes requested):

src/debsbom/generate/generate.py
hakandilek: Alright, that'll make things a lot easier.

.github/actions/pytest/action.yml
hakandilek: reverted this one

src/debsbom/generate/generate.py
hakandilek: Yes, added above the line

src/debsbom/generate/generate.py
hakandilek: simplified the implementation

Urist-McGit (approved):

PR #258, opened 2026-08-13
feat(generate): accept tar root filesystems

### LINKED ISSUE #236: Support scanning tarballs
**Use case:** I want to use the prebuilt debsbom container to create an SBOM for another container.

The [example](https://github.com/siemens/debsbom/blob/69eed5b4e1767bbaa85158760e061e75e494edaa/docs/source/examples.rst#container-rootfs-using-podman) does not seem to support this because I would need the mount point inside the debsbom container. As a workaround, I run `podman export $CRT -o rootfs.tar`, unpack `rootfs.tar` (which messes up file permissions), and mount that directory into debsbom. It works, but it would be nicer if I could avoid unpacking the tarball and point debsbom directly at it.

ting-hong-shieh:
Closes #236.

## Problem

`debsbom generate --root` treated every argument as a directory. A container export therefore failed before scanning:

```text
Base: cfc71ed2e868e5fcd196e596fc69abc365b11867
$ debsbom generate -t spdx -r /tmp/debsbom-rootfs-tree.tar
debsbom: error: [Errno 20] Not a directory: '/tmp/debsbom-rootfs-tree.tar/var/lib/dpkg/status'
```

The command also had no way to consume an exported root filesystem from stdin.

## Change

- Accept a root directory, tar archive, or uncompressed tar stream from `--root`.
- Select the external decompressor from `.bz2`, `.gz`, `.xz`, `.zst`, or `.lz4`.
- Materialize only the dpkg, apt, and requested copyright metadata needed by the scanner.
- Extract through `tarfile.extractall()` with a filter that skips members outside that metadata and delegates the rest to `tarfile.data_filter`, which rejects path traversal, links leaving the destination, and special files.
- Leave tarfile's default errorlevel in place: a rootfs may legitimately carry only part of this metadata (stock Debian containers ship no apt lists, `var/lib/dpkg/arch-native` only exists from trixie on), and `generate.py` already handles that downstream. A hard link whose target lies outside the scanned set is skipped with a warning, since it cannot be materialized from the stream.
- Document file-based and stdin-based container export examples.

## Before/after snapshot

Both snapshots used the same archive, created from `tests/root/tree`.

```text
Head: 805ed1b
$ debsbom generate -t spdx -r /tmp/debsbom-rootfs-tree.tar
result: success; 21 packages and 66 relationships
```

With a fixed timestamp, directory, tar-file, and stdin scans produced the same canonical digest, taken over the sorted `(name, versionInfo)` pairs and the sorted `(spdxElementId, relationshipType, relatedSpdxElement)` triples of the SPDX document:

```text
directory  a28318fc00592dda200a06ce3f6f3134fedb52cb956549bcf7a7137cfc491012
tar file   a28318fc00592dda200a06ce3f6f3134fedb52cb956549bcf7a7137cfc491012
stdin      a28318fc00592dda200a06ce3f6f3134fedb52cb956549bcf7a7137cfc491012
```

No external service was called during these scans.

## Verification

- `python -m pytest tests/test_rootfs.py tests/test_generation.py -q` — 51 passed, 4 skipped
- `python -m pytest -q -m 'not online'` — 144 passed, 4 skipped, 25 deselected (the skips need the `zstd` and `lz4` utilities, which are absent locally but present in CI)
- `python -m pytest tests/test_rootfs.py --cov=debsbom.generate.rootfs` — 96% line coverage
- `python -m black --check $(git ls-files '*.py')` — 91 files unchanged
- `sphinx-build -b html docs/source ...` — succeeds on base and head with an identical set of pre-existing warnings
- `git diff --check`

### DIFF
--- a/src/debsbom/generate/rootfs.py
+++ b/src/debsbom/generate/rootfs.py
@@ -1,0 +1,134 @@
+# Copyright (C) 2026 Siemens
+#
+# SPDX-License-Identifier: MIT
+
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
+import io
+import logging
+from pathlib import Path, PurePosixPath
+import shutil
+import subprocess
+import tarfile
+from tempfile import TemporaryDirectory
+
+from ..util.compression import Compression, CompressionToolMissing
+
+logger = logging.getLogger(__name__)
+
+_ROOT_FILES = frozenset(
+    {
+        PurePosixPath("var/lib/dpkg/status"),
+        PurePosixPath("var/lib/dpkg/arch-native"),
+        PurePosixPath("var/lib/apt/extended_states"),
+    }
+)
+_APT_LISTS = PurePosixPath("var/lib/apt/lists")
+_COPYRIGHT_ROOT = PurePosixPath("usr/share/doc")
+
+
+def _is_rootfs_metadata(name: str, include_copyright: bool) -> bool:
+    path = PurePosixPath(name.lstrip("/"))
+    if path in _ROOT_FILES:
+        return True
+    if path == _APT_LISTS or _APT_LISTS in path.parents:
+        return True
+    if not include_copyright or _COPYRIGHT_ROOT not in path.parents:
+        return False
+
+    relative_to_doc = path.relative_to(_COPYRIGHT_ROOT)
+    return len(relative_to_doc.parts) == 1 or (
+        len(relative_to_doc.parts) == 2 and relative_to_doc.name == "copyright"
+    )
+
+
+def _metadata_filter(
+    include_copyright: bool,
+) -> Callable[[tarfile.TarInfo, str | Path], tarfile.TarInfo | None]:
+    """Build an extraction filter selecting rootfs metadata, hardened by :data:`tarfile.data_filter`.
+
+    Path traversal, links leaving the destination and special files are rejected by the
+    tarfile data filter, which raises a :class:`tarfile.FilterError` for the offending member.
+    """
+
+    def select(member: tarfile.TarInfo, destination: str | Path) -> tarfile.TarInfo | None:
+        if not _is_rootfs_metadata(member.name, include_copyright):
+            return None
+        if member.islnk() and not _is_rootfs_metadata(member.linkname, include_copyright):
+            # the target is skipped, so the link cannot be materialized from the stream
+            logger.warning(
+                f"skipping {member.name}: hard link target '{member.linkname}' "
+                "is outside the scanned rootfs metadata"
+            )
+            return None
+        return tarfile.data_filter(member, destination)
+
+    return select
+
+
+def _materialize_rootfs_metadata(
+    archive: tarfile.TarFile, destination: Path, include_copyright: bool
+) -> None:
+    # A rootfs may legitimately carry only part of this metadata: stock Debian containers
+    # ship no apt lists, and var/lib/dpkg/arch-native only exists from trixie on. Missing
+    # data is handled downstream, so tarfile's default errorlevel is kept and only the
+    # filter rejections below stay fatal.
+    archive.extractall(path=destination, filter=_metadata_filter(include_copyright))
+
+
+@contextmanager
+def _decompressed_tar_stream(path: Path) -> Iterator[io.BufferedIOBase]:
+    try:
+        compression = Compression.from_ext(path.suffix)
+    except ValueError:
+        compression = Compression.NONE
+
+    if compression == Compression.NONE:
+        with path.open("rb") as stream:
+            yield stream
+        return
+
+    tool = shutil.which(compression.tool)
+    if tool is None:
+        raise CompressionToolMissing(compression.tool)
+    process = subprocess.Popen(
+        [tool] + compression.extract + [path],
+        stdout=subprocess.PIPE,
+    )
+    assert process.stdout is not None
+    try:
+        yield process.stdout
+    finally:
+        process.stdout.close()
+        return_code = process.wait()
+    if return_code != 0:
+        raise RuntimeError(f"decompression of {path} failed")
+
+
+@contextmanager
+def rootfs_directory(
+    root: str | Path,
+    input_stream: io.IOBase | None = None,
+    include_copyright: bool = False,
+) -> Iterator[Path]:
+    """Yield a scan-ready directory for a directory, tar file, or tar stream rootfs."""
+    if str(root) != "-":
+        root_path = Path(root)
+        if root_path.is_dir():
+            yield root_path
+            return
+        if not root_path.is_file():
+            raise FileNotFoundError(f"rootfs path does not exist: {root_path}")
+
+    with TemporaryDirectory(prefix="debsbom-rootfs-") as temp_dir:
+        materialized_root = Path(temp_dir)
+        if str(root) == "-":
+            if input_stream is None:
+                raise ValueError("rootfs stdin input requires a binary stream")
+            with tarfile.open(fileobj=input_stream, mode="r|*") as archive:
+                _materialize_rootfs_metadata(archive, materialized_root, include_copyright)
+        else:
+            with _decompressed_tar_stream(Path(root)) as stream:
+                with tarfile.open(fileobj=stream, mode="r|") as archive:
+                    _materialize_rootfs_metadata(archive, materialized_root, include_copyright)
+        yield materialized_root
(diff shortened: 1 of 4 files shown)


src/debsbom/generate/rootfs.py
```diff
@@ -0,0 +1,166 @@
...
+            output_path.parent.mkdir(parents=True, exist_ok=True)
+            relative_target = os.path.relpath(
+                destination.joinpath(*target_path.parts), output_path.parent
+            )
+            output_path.symlink_to(relative_target)
+            continue
+        if not member.isfile():
+            raise ValueError(f"unsupported tar archive member type: {member.name}")
+
+        source = archive.extractfile(member)
```
fmoessbauer: Hi, I'm wondering if we could just use `.extractall()` here (with the expected paths) and use the `filter='data'` to harden against path injections. The manual sanitizing of paths is quite error prone and gives a false sense of safety.

By delegating it to the tar library, we also shift the security boundary towards them and profit from security fixes in the future.

ting-hong-shieh: You're right — that normalization was duplicating `tarfile.data_filter`, and writing the replacement turned up two bugs in the hand-rolled version. Fixed in b2d3f3d.

Extraction now goes through `extractall()` with a filter that returns `None` for members outside the scanned metadata and delegates everything else to `data_filter`. Both path helpers are gone, and it works in the streaming (`r|*`) path too.

Two things the switch surfaced, now covered by tests:

- tarfile swallows non-fatal extraction errors by default, so a hard link to a member we deliberately skip would drop metadata from the SBOM with no diagnostic at all. The archive is extracted with `errorlevel = 2` now, and that hard link is rejected with an explicit message instead of a bare `could not change mode`.
- Members outside the scanned metadata are ignored rather than aborting the scan. That also drops the old blanket rejection of any member name containing a backslash, which could fail a container export over a file we never read.

One hand-rolled check stayed: the hard link case above. It is not a path-safety check — without it the metadata silently vanishes — but say the word and I'll drop it so the filter is nothing but `data_filter`.

Validation:

```text
pytest -q -m 'not online'                              143 passed, 4 skipped, 25 deselected
pytest tests/test_rootfs.py --cov=debsbom.generate.rootfs   96%
```

Directory, tar file and stdin scans of the same rootfs still agree at 21 packages and 66 relationships.

Worth your call: `filter=` landed in 3.11.4 and the project declares `requires-python >= 3.11`. I left it unguarded. I can also squash this into the feature commit before merge if you prefer a single commit.

src/debsbom/generate/rootfs.py
```diff
@@ -0,0 +1,127 @@
...
+
+    return select
+
+
+def _materialize_rootfs_metadata(
+    archive: tarfile.TarFile, destination: Path, include_copyright: bool
+) -> None:
+    # an SBOM must not silently lose metadata, so extraction errors are fatal
+    archive.errorlevel = 2
+    archive.extractall(path=destination, filter=_metadata_filter(include_copyright))
```
fmoessbauer, line 76: We don't necessarily have all information. For example, in off-the-shelf debian containers the `/var/lib/apt/lists/` (apt-cache) is missing. Also `/var/lib/dpkg/arch-native` is only available from debian trixie on. Not having this data should not result in errors. I rather prefer to simply ignore this at ingress / tar extraction and let debsbom internally handle the missing data (similar to manually extracting the rootfs and passing the result to `debsbom -r`).

tests/test_rootfs.py
```diff
@@ -0,0 +1,212 @@
...
+        assert not any(root.iterdir())
+
+
+def test_tar_rootfs_keeps_symlinks_to_unmaterialized_paths_inside_the_root():
+    member = tarfile.TarInfo("var/lib/apt/lists/alias")
+    member.type = tarfile.SYMTYPE
+    member.linkname = "../../dpkg/status"
+
+    with rootfs_directory("-", _stdin_rootfs(member)) as root:
+        link = root / "var/lib/apt/lists/alias"
```
fmoessbauer, line 228: I'm wondering if we should just reject symlinks in the files we process. AFAIK debian itself does not use any, so why even allow it at all?

src/debsbom/generate/rootfs.py
```diff
@@ -0,0 +1,127 @@
...
+
+from ..util.compression import Compression, CompressionToolMissing
+
+_ROOT_FILES = {
+    "var/lib/dpkg/status",
+    "var/lib/dpkg/arch-native",
+    "var/lib/apt/extended_states",
+}
+_APT_LISTS = "var/lib/apt/lists"
+_COPYRIGHT_ROOT = PurePosixPath("usr/share/doc")
```
fmoessbauer, line 27: This is inconsistent with the other variables. Either always create them as `Path`, or just use strings and convert to path on usage.

fmoessbauer:
@ting-hong-shieh Your changes already look pretty good. Please just address the remaining comments and squash both commits into a single one (and then force-push).

fmoessbauer:
@ting-hong-shieh : Any news on this MR? If you like, I can also take over and apply the last remaining fixes.

src/debsbom/generate/rootfs.py
ting-hong-shieh: Changed — the strict posture is gone. `errorlevel = 2` and the comment above it are
removed, so tarfile's default applies: filter rejections (traversal, links leaving the
destination, special files) stay fatal, everything else is non-fatal. The hard link case
below is now skipped with a warning instead of raising.

One correction to my own change though: I could not reproduce the failure as described.
A rootfs simply lacking `var/lib/apt/lists/` or `var/lib/dpkg/arch-native` already
extracted and scanned fine before this commit — those members are just absent from the
archive, so nothing raises, and `generate.py` already falls back (`_derive_distro_arch`,
the `Missing apt extended_states file` info log, `except FileNotFoundError` around
copyright). Verified against a tar of `tests/root/arch-detect`, which has only
`var/lib/dpkg/status`:

```
$ debsbom generate -t spdx -o out --root partial.tar
INFO:debsbom.generate.generate:Missing apt extended_states file, all packages will be marked as manually installed
...exit 0
```

So what actually changes is the posture you objected to, not a broken case: extraction no
longer claims authority over what counts as complete metadata, and a future non-fatal
tarfile error will not abort a scan. `tests/test_rootfs.py::test_tar_rootfs_tolerates_partial_metadata`
pins the behaviour so it cannot regress.

tests/test_rootfs.py
ting-hong-shieh: I would rather keep them: Debian does use symlinks in exactly the tree we read for
licenses. On a stock Ubuntu 24.04 rootfs, 43 of the 627 entries directly under
`/usr/share/doc` are symlinks to sibling directories, and copyright files themselves can
be symlinks too:

```
/usr/share/doc/libstdc++6      -> gcc-14-base
/usr/share/doc/openssl/copyright -> ../libssl3t64/copyright
```

This is the shared-copyright convention: when several binaries come from one source
package, only one directory carries the real `copyright` and the rest link to it.
Rejecting symlinks would silently drop license data for those packages, which is the
failure mode this PR is trying to avoid.

They stay hardened either way — `tarfile.data_filter` rejects any link resolving outside
the destination, covered by `test_tar_rootfs_rejects_metadata_symlink_escape` and
`test_copyright_directory_symlink_stays_inside_materialized_root`.

Happy to reject them under `var/lib/dpkg` and `var/lib/apt` specifically if you want the
narrower rule — Debian genuinely does not use symlinks there, and that would not cost any
license data.

src/debsbom/generate/rootfs.py
ting-hong-shieh: Done — all three are `PurePosixPath` now, and the matching uses pathlib rather than string
prefixes:

```python
if path in _ROOT_FILES:
    return True
if path == _APT_LISTS or _APT_LISTS in path.parents:
    return True
if not include_copyright or _COPYRIGHT_ROOT not in path.parents:
    return False
relative_to_doc = path.relative_to(_COPYRIGHT_ROOT)
```

tests/test_rootfs.py
fmoessbauer: Ah... right. Then let's keep the logic simple and just accept them.

fmoessbauer (approved):
Thanks for contributing this feature!

src/debsbom/generate/rootfs.py
fmoessbauer: The filter parameter actually landed in `3.12`, not `3.11.4`. By that, we either need to bump the python version or (preferable) make the filter optional and issue a warning if not available. We still need to support debian bookworm, so bumping to `3.12` is not really an option.

PR #260, opened 2026-08-20
Various non-critical hardenings against user errors and malicious inputs

fmoessbauer:
For details, see the individual commit messages.

### DIFF
--- a/src/debsbom/snapshot/client.py
+++ b/src/debsbom/snapshot/client.py
@@ -12,6 +12,8 @@ from dataclasses import dataclass
 from datetime import datetime
 from debian import deb822
 import logging
+import posixpath
+import re
 import requests
 from requests.exceptions import RequestException
 
@@ -29,6 +31,16 @@ logger = logging.getLogger(__name__)
 
 
 UPSTREAM_ARCHIVE_ORDER = ["debian", "debian-security", "debian-debug", "debian-ports"]
+
+# Characters allowed in a debian package file name: package name, version
+# (incl. epoch ':' and revision '-'), architecture, separators and extension.
+_FILENAME_RE = re.compile(r"[A-Za-z0-9][A-Za-z0-9.+~:_-]*")
+
+# A SHA-1 digest as used to address files on the snapshot farm.
+_SHA1_RE = re.compile(r"[0-9a-fA-F]{40}")
+
+# Characters allowed in a debian archive name.
+_ARCHIVE_RE = re.compile(r"[a-zA-Z0-9-]+")
 
 
 class SnapshotDataLakeError(Exception):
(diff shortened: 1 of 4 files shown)


docs/source/intro.rst
```diff
@@ -88,3 +88,11 @@ Reliably and correctly identifying if a package is a vendor package or not is no
...
 
 .. [#f1] https://github.com/CycloneDX/specification/issues/612
+
+Security Boundaries
+-------------------
+
+Unless stated otherwise, all input to debsbom is considered trusted.
+Data from external services (e.g. the Debian snapshot service) is sanitized on a best-effort basis: its meta-data is untrusted and sanitized by debsbom, while the file content itself is considered trusted.
+If an SBOM provides checksums for an artifact (e.g. a source or binary package), the downloaded artifact is verified against them, including transitive checksums such as those in a ``.dsc`` file; without checksums, artifacts are considered trusted and not verified.
+All debsbom plugins are considered trusted code.
```
jan-kiszka: There is no SECURITY.md in the project yet. Wouldn't it make more sense to add that, link to this from there or even include this section there?

jan-kiszka (commented):
Looks generally good. For patch 1, I'm missing an impact explanation (what was possible so far?).

docs/source/intro.rst
fmoessbauer: Yes, but I would like to clarify this with the co-maintainers before creating one. For now, we just add it to the documentation to have it at least informally stated. I will create an issue for moving this.

fmoessbauer:
> Looks generally good. For patch 1, I'm missing an impact explanation (what was possible so far?).

After double checking the code, I noticed that the path traversal bug was already fixed in 69eed5b4e. We now just protect debsbom plugins by sanitizing upfront.

By that, this PR does only contains hardenings, but no security fixes.

docs/source/intro.rst
fmoessbauer: I prepared one and updated the commit. Let's wait for Christoph to hear his opinion as well.

src/debsbom/snapshot/client.py
```diff
@@ -177,19 +188,62 @@ class SnapshotRemoteFile(RemoteFile):
...
+        """
+        Ensure the archive name only consists of the allowed set of characters
+        and return it. Raises ValueError otherwise.
+        """
+        if not _ARCHIVE_RE.fullmatch(name):
+            raise ValueError(f"invalid archive name: {name!r}")
+        return name
+
+    @staticmethod
+    def _sanitize_path(path: str) -> str:
```
Urist-McGit, line 223: This might reject valid paths. A path like path/to/../xyz would be rejected even though it is unproblematic, if unconventional.

Urist-McGit (commented):
The SECURITY.md looks fine to me. This would create a discrepancy from the group provided security policy as it is displayed on GitHub, but I think we do not need to adhere to it anyways.

jan-kiszka:
The purpose of the generic policy is to provide a fallback if the project itself does not provide a more specific policy. And that is exactly what is done here. Or do you see specific aspects where we should continue to follow the Siemens policy?

src/debsbom/snapshot/client.py
fmoessbauer: Fixed.

Urist-McGit (approved):

PR #263, opened 2026-08-27
Add support to select CycloneDX schema version

### LINKED ISSUE #261: Update default cycloneDX output to 1.7
Hello,

I'd like to be able to output SBOMs using the latest [cyclone-dx version - 1.7](https://github.com/CycloneDX/specification/releases#release-1.7).  Currently the version number is fixed at 1.6 in [cdxbomwriter](https://github.com/siemens/debsbom/blob/main/src/debsbom/bomwriter/cdxbomwriter.py).  Version 1.7 has been the latest release since October of 2025 (about a year ago now).  While the version can be manually updated easily enough in code, it would be great if the user could either select the version (perhaps via a new `--cdx-version` flag) *or* we defaulted to the latest version (simply update the hard code).  

Thanks for your consideration.

fmoessbauer:
Depends on #262 

Closes: #261

### DIFF
--- a/tests/test_generation.py
+++ b/tests/test_generation.py
@@ -18,7 +18,7 @@ from io import TextIOWrapper
 
 from debsbom.apt.cache import ExtendedStates, Repository
 from debsbom.apt import cache as apt_cache
-from debsbom.bomwriter.bomwriter import BomWriter
+from debsbom.bomwriter.bomwriter import BomWriter, SerializerOpts
 from debsbom.dpkg.package import ChecksumAlgo
 from debsbom.util.compression import Compression
 from debsbom.generate import Debsbom, SBOMType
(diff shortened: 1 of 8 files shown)


tests/test_generation.py
```diff
@@ -130,6 +130,25 @@ def test_default_spdx_creation_timestamp_has_timezone():
...
+    import cyclonedx.schema as cdx_schema
+
+    opts = SerializerOpts.create(SBOMType.CycloneDX, schema_version)
+
+    expected = (
+        max(cdx_schema.SchemaVersion)
+        if expected_schema_version is None
+        else cdx_schema.SchemaVersion.from_version(expected_schema_version)
+    )
+    assert opts.cdx_schema_version == expected
```
Urist-McGit, line 149: This test will fail if the cyclonedx lib gets updated to include a new latest schema version. We have no upper bound for it specified as of right now.

Urist-McGit:
Also needs a rebase as the first couple of commits have been merged already with #262

fmoessbauer:
@Urist-McGit rebase done

Urist-McGit (changes requested):

tests/test_generation.py
fmoessbauer: Yes, indeed. This test makes no sense. I'll drop it.

fmoessbauer: Hmm. I just read the code again. The test is correct. We don't specify or check for a upper bound.

Urist-McGit (approved):

PR #266, opened 2026-09-01
fix(generate): Extract tar without filter if tarfile is too old

fmoessbauer:
The tarfile data_filter was added in 3.12, but Debian bookworm uses 3.11. As we want to keep support for running debsbom on Debian bookworm (e.g. in the isar build system), we implement a fallback. In case tarfile does not offer the option, we simply extract the whole rootfs and issue a warning.

Fixes: f08adbe ("feat(generate): accept tar root filesystems")

### DIFF
--- a/src/debsbom/generate/rootfs.py
+++ b/src/debsbom/generate/rootfs.py
@@ -73,7 +73,15 @@ def _materialize_rootfs_metadata(
     # ship no apt lists, and var/lib/dpkg/arch-native only exists from trixie on. Missing
     # data is handled downstream, so tarfile's default errorlevel is kept and only the
     # filter rejections below stay fatal.
-    archive.extractall(path=destination, filter=_metadata_filter(include_copyright))
+    if hasattr(tarfile, "data_filter"):
+        archive.extractall(path=destination, filter=_metadata_filter(include_copyright))
+    else:
+        logger.warning(
+            "tarfile.data_filter not available, extracting all rootfs metadata without filtering. "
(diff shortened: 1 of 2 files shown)


src/debsbom/generate/rootfs.py
```diff
@@ -73,7 +73,13 @@ def _materialize_rootfs_metadata(
     # ship no apt lists, and var/lib/dpkg/arch-native only exists from trixie on. Missing
     # data is handled downstream, so tarfile's default errorlevel is kept and only the
     # filter rejections below stay fatal.
-    archive.extractall(path=destination, filter=_metadata_filter(include_copyright))
+    if hasattr(tarfile, "data_filter"):
+        archive.extractall(path=destination, filter=_metadata_filter(include_copyright))
+    else:
+        logger.warning(
+            "tarfile.data_filter not available, extracting all rootfs metadata without filtering"
```
Urist-McGit: Maybe the warning should give a summary of what the big problem is, and how to resolve it. For example: `tarfile.data_filter not available, extracting all rootfs metadata without filtering. This is a potentially unsafe operation. To mitigate the risk please update your Python version to >3.11.4.`

Urist-McGit (changes requested):

src/debsbom/generate/rootfs.py
fmoessbauer: Done

Urist-McGit (approved):

PR #268, opened 2026-09-03
Prepare v0.10.1 release

Urist-McGit:
Proposed changelog:

# Bug fixes

- Fix crash when generating from a tarball and using Python version <3.11.4
  Some filtering features we use for security hardening were added in this patch release. If the Python version is older we fall back on a less performant and potentially unsafe extraction method. A warning is emitted in this case.


fmoessbauer (approved):
Tested in isar.

PR #269, opened 2026-09-04
Fix missing distro qualifier for specific apt cache situations

Urist-McGit:
For cases where apt cache binary package data is available, but not the source package data we missed distro qualifiers for source packages derived from these binary packages (Source, Built-Using, Static-Built-Using).

### DIFF
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -228,7 +228,7 @@ class Package(ABC):
         """
         logger.info(f"Parsing status file '{status_file}'...")
         deb822_stream = open(status_file, "r", encoding="utf-8")
-        pkgs_it = cls.inject_src_packages(filter_installed(cls._parse_dpkg_status(deb822_stream)))
+        pkgs_it = filter_installed(cls._parse_dpkg_status(deb822_stream))
         return PkgListStream(deb822_stream, PkgListType.STATUS_FILE, pkgs_it)
 
     @classmethod
--- a/src/debsbom/dpkg/package.py
+++ b/src/debsbom/dpkg/package.py
@@ -402,10 +402,10 @@ class Package(ABC):
             # Add partial source package. This can later be enhanced by merging it with
             # a more complete source package we discovered via other mechanisms (e.g. apt cache).
             logger.debug(f"Found built-using source package: '{bu.name}@{bu.version[1]}'")
-            yield SourcePackage(bu.name, bu.version[1], pkg.distro)
+            yield SourcePackage(bu.name, bu.version[1], distro=pkg.distro)
         for sbu in pkg.static_built_using:
             logger.debug(f"Found static-built-using source package: '{sbu.name}@{sbu.version[1]}'")
-            yield SourcePackage(sbu.name, sbu.version[1], pkg.distro)
+            yield SourcePackage(sbu.name, sbu.version[1], distro=pkg.distro)
         if add_pkg:
             yield pkg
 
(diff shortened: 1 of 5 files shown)


tests/test_generation.py
```diff
@@ -940,19 +940,12 @@ def test_apt_codename(tmpdir, sbom_generator):
             for ref in package["externalRefs"]:
                 if ref["referenceType"] == "purl":
                     seen = True
-                    # only the debsbom packages are referenced in the apt cache
-                    if "debsbom" in package["name"]:
-                        assert "distro=codename-stable" in ref["referenceLocator"]
-                    else:
-                        assert "distro=codename-stable" not in ref["referenceLocator"]
+                    assert "distro=codename-stable" in ref["referenceLocator"]
```
fmoessbauer, line 943: Please move to standalone commit, as the other commits are fixes.

src/debsbom/dpkg/package.py
```diff
@@ -402,10 +402,10 @@ def _resolve_sources(cls, pkg: "BinaryPackage", add_pkg=False) -> Iterable["Pack
             # Add partial source package. This can later be enhanced by merging it with
             # a more complete source package we discovered via other mechanisms (e.g. apt cache).
             logger.debug(f"Found built-using source package: '{bu.name}@{bu.version[1]}'")
-            yield SourcePackage(bu.name, bu.version[1], pkg.distro)
+            yield SourcePackage(bu.name, bu.version[1], distro=pkg.distro)
```
fmoessbauer, line 405: Good catch. Do you have an estimate how severe this bug was in practice? Does this qualify for a 0.10.2 release?

Urist-McGit: It depends on the exact availability of the apt cache, but it usually only means missing distro informatiion and a wrong maintainer for built-using packages.

I actually intended a patch release for this. We currently inject source packages two times, but removing the first injection from the parsing of dpkg status file would be a breaking APi change, which I think we should include at the next major/minor release.

fmoessbauer: > removing the first injection from the parsing of dpkg status file would be a breaking APi change, which I think we should include at the next major/minor release.

Why? Shouldn't the SBOM be identical in almost all cases? This is not a user facing change - or is it?

tests/test_generation.py
Urist-McGit: Ah wait, this would lead to failing tests when it is not done here. The test incorrectly assumed that all packages that are not in the apt cache should not get the distro qualifier, but that is no longer true since a343ed57a3eb5b01c1cffcead621c396a5e4c49d (even if that commit had some bugs and didnt work)

src/debsbom/dpkg/package.py
Urist-McGit: If you just parse a dpgk status file with our API you would be missing all source packages which were previously included.

tests/test_generation.py
fmoessbauer: Okay, but that means that the test was basically wrong in the first place (even prior to a343ed5), and a343ed5 should have exposed this but did not because it itself was buggy. Well... #254 is not a success story :)

Then let's keep it like that but mention it in the commit message.

Urist-McGit: Agreed, I will mention the test changes in the message

src/debsbom/dpkg/package.py
fmoessbauer: The API itself is not part of the public interface (except for the plugins API). We even have a disclaimer for that. By that, we don't have to care and better just fix this for the CLI users in a bugfix release.

> The API is not stable (yet) and by that should not be imported into third party modules.

fmoessbauer (approved):

PR #271, opened 2026-09-17
Add --artifact flag to generate command to associate a file with SBOM

Urist-McGit:
By associating a file artifact with the generated SBOM we can provide artifact related information for the root component.

The filename is referenced the packageFileName field in SPDX, and ~as external file:// reference in CDX~ in `evidence.occurrences`. Hashes of the artifact are also calculated and added to the root component, along the omnibor artifact ID and the swhid.

The components keep their type as operating system, even with an associated artifact. The CDX spec includes operating  system images and there is no detailed description in SPDX, but we assume it is similar.

[1] https://omnibor.io/docs/artifact-ids/


src/debsbom/util/omnibor.py
```diff
@@ -0,0 +1,25 @@
...
+    The artifact should be in uncompressed form as otherwise we depend
+    on the compression algorithms reproducability guarantees.
+    """
+
+    hasher = sha256()
+    with open(artifact, "rb") as f:
+        size = os.fstat(f.fileno()).st_size
+
+        hasher.update(f"blob {size}\0".encode("ascii"))
+        hasher.update(f.read())
```
fmoessbauer: This doesn't work with large files, as it does not read the whole file, but just a single chunk. We either need an approach like https://github.com/siemens/kas/blob/055518846c3963114efbd0107bf3f2e59b4b3ba5/kas/attestation.py#L52 or maybe the hasher has built-in support for adding a file.

tests/test_misc.py
```diff
@@ -0,0 +1,16 @@
...
+from pathlib import Path
+
+
+def test_artifact_id():
+    artifact = Path("tests/data/omnibor.txt")
+
+    artifact_id = omnibor_artifact_id(artifact)
+    assert (
+        artifact_id
+        == "gitoid:blob:sha256:cb16a7604bae14bc2d888df559984c2c60920a65c8f4645a7583aa0f1dee8341"
```
fmoessbauer, line 17: Did you compare the hash with other implementations of omnibor? How can we be sure that the hash is computed correctly?

src/debsbom/util/omnibor.py
Urist-McGit: Unfortunately there is no built-in support to update a hasher with a file in the stdlib module. I will copy the approach from kas.

src/debsbom/generate/spdx.py
```diff
@@ -54,6 +56,16 @@ def make_distro_package(
             name=distro_supplier,
         )
 
+    external_references = None
+    if distro_omnibor_id:
```
fmoessbauer: Can / should we store the original filename in the sbom as well? Also, shall we add the other hashes we compute for components also here (like sha256, sha1, ...)?

tests/test_misc.py
Urist-McGit: I used omnibor-rs to generate the reference value

src/debsbom/generate/spdx.py
Urist-McGit: I am not sure where exactly to include the filename in the SBOM.

We could include checksums for the artifact just like we provide the omnibor ID. The `--artifact` flag is generic enough to include any data about the artifact in the root component.

fmoessbauer: This brings us back to the original issue: What does an omnibor / hash refer to? As of now, this cannot be inferred from the SBOM. We don't even know if it refers to a compressed artifact or the extracted one. How do others encode this information?

tests/test_misc.py
fmoessbauer: Maybe add this info as a comment.

src/debsbom/generate/spdx.py
Urist-McGit: Thinking about it we would need to create a new component in the SBOM that represents the associated file itself.

fmoessbauer: That IMHO makes sense. This node could even be used as root node.

Urist-McGit: According to the CDX spec (bold by me):
> "operating-system"	
A software operating system without regard to deployment model (i.e. installed on physical hardware, virtual machine, **image**, etc) Refer to https://en.wikipedia.org/wiki/Operating_system.


we could simply reference the filename as external reference and include the hashes as I did already.

For SPDX there is no further explanation what an operating system is exactly, but we could similarly just assume the same as for CDX and provide the filename in the root component.

src/debsbom/util/swh.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def swh_id(artifact: Path) -> str:
+    """
+    Calculate the content swhid for the given artifact.
+
+    The artifact should be in uncompressed form as otherwise we depend
+    on the compression algorithms reproducability guarantees.
+    """
+    digest = gitoid_hash(artifact)
```
fmoessbauer, line 14: Now we hash twice, despite it uses exactly the same hash. Please only hash once.

src/debsbom/util/omnibor.py
```diff
@@ -0,0 +1,25 @@
...
+
+from .gitoid import gitoid_hash
+
+
+def artifact_id(artifact: Path) -> str:
+    """
+    Calculate the omnibor artifact ID for the given artifact.
+
+    The artifact should be in uncompressed form as otherwise we depend
+    on the compression algorithms reproducability guarantees.
```
fmoessbauer: This does not matter here, as we just hash what the user provides. But we need to document this recommendation in the CLI docs.

src/debsbom/util/omnibor.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    digest = gitoid_hash(artifact)
+    return artifact_id_from_digest(digest)
+
+
+def artifact_id_from_digest(digest: str) -> str:
```
fmoessbauer: Be precise about the digest and either call this digest_sha256 or provide a annotated hash - similar to how we handle hashes in other places.

src/debsbom/generate/generate.py
```diff
@@ -354,6 +356,7 @@ def generate(
         """
         Generate SBOMs. The progress callback is of format: (i,n,package)
         """
+
```
fmoessbauer: Stray newline

Urist-McGit:
Added documentation and tests. Of course after adding the tests some things needed adjustements: swhid is now correctly generated with SHA1 instead of SHA256.

Also moved the CycloneDX filename reference to `evidence.occurrences` instead of external references.

src/debsbom/util/gitoid.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (C) 2026 Siemens
+#
+# SPDX-License-Identifier: MIT
+
+from hashlib import sha256
+import os
+from pathlib import Path
+
+
+def gitoid_hashes(artifact: Path, hashers=None) -> list[str]:
```
fmoessbauer, line 10: Nit: Why not type annotation for the hashers?

Urist-McGit: I am not quite sure I can reference the used class:
```
>>> type(sha256())
<class '_hashlib.HASH'>
>>> type(sha1())
<class '_hashlib.HASH'>

```

fmoessbauer: Ok, then lets keep it as is.

fmoessbauer (approved):

PR #272, opened 2026-09-17
filter: support source-aware CycloneDX package exclusions

### LINKED ISSUE #259: Support package exclusions when filtering CycloneDX SBOMs
### Problem

Consumers that need to omit installed Debian packages currently have to post-process the generated CycloneDX document. `debsbom filter` can retain binaries, sources, or one named package, but it cannot consume an exclusion expression or a source-package/version exclusion file. `download --skip-pkgs` accepts package-list content but does not remove components from the SBOM.

This makes wrappers responsible for parsing installed source-to-binary relationships, pruning components before source acquisition, and reproducing `CdxSbomFilter` graph cleanup.

### Proposed capability

Add native filtering inputs that can:

- exclude binary packages by a validated pattern matched against package identity;
- exclude source name/version entries and expand them to all related installed binaries;
- operate before `download` and `repack`, so excluded sources are not acquired;
- remove excluded component nodes, their dependency entries, and every reference to them;
- retain a source while any retained binary was produced by it or still references it through `Built-Using` / `Static-Built-Using`;
- support a valid empty result;
- fail explicitly when a requested installed package cannot be mapped to a generated component.

A structured exact-name/version interface would be preferable to requiring every consumer to reproduce legacy scanner regex syntax.

### Wrapper-removal condition

The deb-scanner wrapper can delete its CycloneDX package post-processing once a released debsbom version provides both exclusion forms above, performs graph-consistent source-aware pruning before acquisition, and exposes enough result information to report matched, removed, and unmatched exclusions.

Shubham-Padkonde:
Consumers currently have to remove packages from generated CycloneDX inventories in their own wrappers. This adds graph-based exclusions to `filter` so the resulting inventory can be passed to `download` and `repack`.

Fixes #259.

- Repeatable `--exclude-binary REGEX` matches entire binary package names.
- `--exclude-source-file FILE` reads exact source name/version pairs using the existing universal ingress formats: package lists, Debian PURLs, Isar manifests, and dpkg status files. Only source entries are selected. Excluding a source preserves its binaries and other consumers.
- Matching packages and their references are removed, then newly unreferenced dependencies are pruned. Packages still referenced by retained packages are preserved.
- The existing global `--json` flag writes a single-line exclusion report to stdout. A separate SBOM output file is required to avoid mixing the report with the SBOM.
- The JSON report schema is included. Exclusions run before `--sources`/`--binaries`; combining exclusions with `--package` or SPDX input is rejected. SPDX exclusion support remains a follow-up.

Source input now uses the existing universal ingress parser as requested in review; the custom JSONL input schema has been removed.

Validation of the revised code (Windows, Python 3.13): all 35 filter tests pass. Black leaves the three changed Python files unchanged (with a Python target-version warning), and `git diff --check` passes. The broader offline suite stopped at three failures: 130 passed, 10 skipped, 25 deselected. Failures were local download URL handling, local apt-cache parsing, and a rootfs symlink assertion. The revised full Linux suite has not been run; this is not a claim of fully passing CI.

The contributor personally signed off the implementation and follow-up commits. Implementation and validation were assisted by Codex and Claude.

Latest universal-ingress validation (Windows, Python 3.13): all four input-format regression cases failed against the old reader; all 35 filter tests pass after the change. In an SPDX-only environment, 8 pass and 27 are skipped because CycloneDX is absent. Black and git diff --check pass.


Urist-McGit:
Thanks for the contribution! We will have a look in the next days, if nothing happens until end of next week dont be afraid to ping us again ;)

docs/source/commands/filter.rst
```diff
@@ -8,3 +8,62 @@
...
+
+For exact source-package exclusions, provide a JSON array of source names and
+Debian versions using ``--exclude-source-file``. For example:
+
+.. code-block:: json
+
+    [
+      {"name": "linux", "version": "6.12.73-1"},
+      {"name": "example-source", "version": "2:1.0-1"}
+    ]
```
fmoessbauer, line 35: We need a json schema for this (similar to the other schemas we already have). I'm also wondering if it would be better to ingest this in json-lines format instead of the array. That's easier to create and "compose".

docs/source/commands/filter.rst
```diff
@@ -8,3 +8,62 @@
...
+
+Source exclusions require the **matching system's** dpkg status file::
+
+    debsbom filter input.cdx.json filtered.cdx.json \
+        --exclude-source-file sources.json \
+        --installed-status rootfs/var/lib/dpkg/status \
+        --exclusion-report exclusions.json
```
fmoessbauer, line 42: We also need a json schema for the report.

src/debsbom/commands/filter.py
```diff
@@ -74,6 +133,8 @@ def get_package_repr(sbom_type: SBOMType) -> tuple[PackageRepr, str]:
             SbomOutput.write_out_arg(
                 resolver.document, resolver.sbom_type(), args.bomout, args.validate
             )
+            if report_file:
+                Path(report_file).write_text(json.dumps(report, indent=2) + "\n", encoding="utf-8")
```
fmoessbauer, line 137: AFAIK we don't indent the json outputs at all (except for the sboms). We should do this consistently across debsbom.

src/debsbom/commands/filter.py
```diff
@@ -95,3 +156,31 @@ def setup_parser(cls, parser):
             type=str,
             help="filter the SBOM by only including the package and its dependency subgraph",
         )
+
+        parser.add_argument(
+            "--exclude-binary",
+            action="append",
+            metavar="REGEX",
```
fmoessbauer, line 163: Please use the same metavar as the argument (just don't provide it). Otherwise it is likely that this collides with other argument vars.

fmoessbauer: Ah... sorry. Metavar is for the placeholder name in the docs. Forget what I said ...

src/debsbom/filter/cdx.py
```diff
@@ -62,3 +65,152 @@ def packages(cls, document: Bom, source_pkg: PackageRepr, packages: list[Package
...
+        for pattern in patterns:
+            try:
+                compiled.append(re.compile(pattern))
+            except re.error as error:
+                raise ValueError(
+                    f"invalid binary exclusion pattern {pattern!r}: {error}"
+                ) from error
+        sources = list(dict.fromkeys(source_packages or []))
+        for name, version in sources:
+            if not re.fullmatch(r"[a-z0-9][a-z0-9+.-]*", name) or not version:
```
fmoessbauer, line 91: Why compile the other regexes, but not this one? Please compile all regex.

Urist-McGit:
> An explicit --installed-status file maps them to installed binaries without confusing a primary source with Built-Using dependencies.

This is certainly not necessary.

In general I think the approach is not the right one: you go from the actual package representations and figure out the to be removed dependencies from there. This should not be necessary, you can simply stay in the graph domain of the SBOM. I would suggest something like this:
1. figure out which packages match the passed in regular expressions
2. remove packages and all their references from the dependency graph
3. prune the graph by removing any unreferenced packages
4. repeat step 2 and 3 until no more packages are removed in step 3

That way you remove all dependencies that are not also referenced by a package that was not removed, and do not need to have any actual packages to deal with.

Urist-McGit:
We would like to have support for SPDX SBOMs too, even if the issue does not mention it. But I can see that this more or less doubles the implementation effort, so this is not blocking to get this merged eventually.

Urist-McGit:
There is a test failure that is related to a missing extra for the CDX library we use. If I understand correctly you validate the SBOM there. The validation this way should not be needed, there is a validator already included for the internal representation. It should also be available to use in the tests.

Shubham-Padkonde:
Thanks for the reviews. I reworked this in aa9a5ce along the suggested graph-based approach:

- Matching packages (binary regex, exact source name/version) are removed with their references, then unreferenced packages are pruned until nothing changes. `--installed-status` is gone. The binaries of an excluded source are determined with the existing binary -> source mapping of `CdxPackageResolver`, since CycloneDX cannot tell a binary's own source apart from Built-Using.
- The source exclusions are now JSON lines, with a schema. The report has its own schema and is written without indentation.
- All regexes are compiled.
- The tests no longer use the CycloneDX JSON validator. The empty result is instead written and read back through debsbom's own reader.

One behaviour to confirm: pruning now also removes binary dependencies that only the excluded packages used (for example a library pulled in only by an excluded package), not just their sources. Packages that were already unreferenced in the input are kept. SPDX is not implemented yet; I'd prefer to do it as a follow-up.

docs/source/commands/filter.rst
Shubham-Padkonde: Done in aa9a5ce: `--exclude-source-file` now reads JSON lines (one `{"name", "version"}` object per line), validated against the new `schema-filter-exclude.json`.

docs/source/commands/filter.rst
Shubham-Padkonde: Done in aa9a5ce: added `schema-filter-report.json`, and the CLI test validates the written report against it.

src/debsbom/commands/filter.py
Shubham-Padkonde: Done in aa9a5ce: the report is now written without indentation, as a single JSON object on one line.

src/debsbom/commands/filter.py
Shubham-Padkonde: Kept the explicit metavar as per your follow-up. It's only the placeholder in the help text.

src/debsbom/filter/cdx.py
Shubham-Padkonde: Done in aa9a5ce: all regexes are now compiled, including the source package name pattern (`SOURCE_NAME_RE`).

src/debsbom/filter/cdx.py
```diff
@@ -62,3 +66,97 @@ def packages(cls, document: Bom, source_pkg: PackageRepr, packages: list[Package
...
+        # binaries built from an excluded source, using the same binary -> source
+        # mapping as the CycloneDX package resolver
+        if matched_sources:
+            resolver = CdxPackageResolver(document)
+            for ref, package in resolver.packages_by_ref.items():
+                if package.is_source() or not package.source:
+                    continue
+                source = (package.source.name, package.source.version[1])
+                if source in selectors:
+                    removed.add(str(ref))
```
Urist-McGit: I dont think we want to remove the attached binary package when we remove the source package. We only want to remove unconnected nodes from the graph.

I think the initial issue said it should be done this way, but I do not agree with that specific point. If only a source package should be removed then we should only remove exactly that. That way the behavior is very simply: only the specified packages are removed, and anything that would be left unconnected from the operation. That is how all other commands work when it comes to remove elements from the SBOM.

docs/source/commands/filter.rst
```diff
@@ -8,3 +8,60 @@
...
+The exclusion works on the dependency graph of the SBOM:
+
+1. the packages matching the exclusions are removed together with all references to them
+2. packages that are no longer referenced by any remaining package are removed
+3. step 2 is repeated until no more packages are removed
+
+This removes the sources and dependencies of the excluded packages unless they are
+still referenced by a retained package, for example through ``Built-Using`` or
+``Static-Built-Using``. Packages that were already unreferenced in the input are
+kept. Excluding all packages produces a valid empty SBOM that retains the root
```
Urist-McGit: This should be an implementation detail and not documented. Unconnected packages are a thing that should never happen under normal circumstances.

docs/source/commands/filter.rst
```diff
@@ -8,3 +8,60 @@
...
+metadata component.
+
+The optional ``--exclusion-report`` is written as a single JSON object to a
+separate file, also when the SBOM is read from or written to standard input/output.
+Unmatched exclusions do not fail the command. The report follows this schema:
```
Urist-McGit: We have a global `--json` flag for other commands. Instead of using this new report mechanic we should reuse this flag for the command too. You can view how it is implemented for the `download` command.

docs/source/commands/filter.rst
```diff
@@ -8,3 +8,60 @@
...
+    debsbom download --sources --outdir downloads filtered.cdx.json
+    debsbom repack --sources --dldir downloads filtered.cdx.json repacked.cdx.json
+
+To exclude source packages by exact name and Debian version, pass a file with one
+JSON object per line to ``--exclude-source-file``:
+
+.. code-block:: json
+
+    {"name": "linux", "version": "6.12.73-1"}
+    {"name": "example-source", "version": "2:1.0-1"}
```
Urist-McGit, line 33: I am a bit unhappy with this json format to specify the source packages. We use the concept of universal ingress in other places to specify packages. Since we are not using regular expressions here we could just the same here. Having a completely different format for specifying binary/source packages also feels wrong, as they are actually very similar, even if we require the specific version instead of just a name for source packages.

We already have a bunch of ways to specify packages, adding more will make it even more confusing for users.

@fmoessbauer any ideas for a good solution?

src/debsbom/commands/filter.py
```diff
@@ -11,13 +14,36 @@
...
 
 class FilterCmd(SbomInput, SourceBinaryInput):
     """Filter SBOMs."""
 
     @classmethod
     def run(cls, args):
         resolvers = cls.get_sbom_resolvers(args)
+        patterns = getattr(args, "exclude_binary", None) or []
+        source_file = getattr(args, "exclude_source_file", None)
+        report_file = getattr(args, "exclusion_report", None)
```
Urist-McGit: Are these `getattr` necessary? If the option was added to the parser it is `None` anyways if not specified.

tests/test_exclusions.py
Urist-McGit, line 1: This should just be part of the `test_filter.py` file.

Urist-McGit (changes requested):
Also please squash your commits when it is just a rework and does not implement a separate logical change. That makes it easier for us to review.

Shubham-Padkonde:
Updated in 0cf0364 and squashed the rework into one personally signed-off commit.

- Source-only exclusions now preserve their binaries and Built-Using consumers.
- Reused global `--json` for the stdout report and removed `--exclusion-report`; report and SBOM output must be separate.
- Removed the redundant `getattr` calls and implementation-detail documentation.
- Moved the exclusion tests into `test_filter.py`, keeping optional CycloneDX imports local so SPDX tests are not skipped with them.

All 35 filter tests pass on Windows/Python 3.13. The broader offline run stopped at three failures (130 passed, 10 skipped, 25 deselected): local download URL handling, local apt-cache parsing, and a rootfs symlink assertion. The PR description records these limits; I am not claiming a passing full Linux suite.

The source-input format remains JSONL while the universal-ingress discussion is pending. Should this accept the existing package-input forms with exact source versions? I have left that review point open rather than introduce a replacement format without agreement.

Prepared with Codex assistance.

Shubham-Padkonde:
The SPDX-only CI failure came from eight new CycloneDX-dependent test functions missing the optional-dependency guard. Added the same pytest.importorskip("cyclonedx") guard used by the existing tests. With CycloneDX installed, all 35 filter tests pass; in an SPDX-only environment, 5 pass and 30 are skipped. Black and git diff --check pass. The contributor personally signed off commit 8b20560. Prepared with Codex assistance.

docs/source/commands/filter.rst
fmoessbauer: The universal ingress format should be good enough. For details, see https://siemens.github.io/debsbom/examples.html#from-package-list

Shubham-Padkonde:
Implemented the universal-ingress request in 376e1c6, personally signed off by the contributor. Source exclusions reuse Package.parse_pkglist_stream for package lists, Debian PURLs, Isar manifests, and dpkg status input. Binary entries are ignored, sources require exact versions, and duplicate source entries are deduplicated. Removed the custom JSONL input schema and updated the help and documentation; the report schema is retained.

All four input-format regressions fail against the old reader. With the change, all 35 filter tests pass; in an SPDX-only environment, 8 pass and 27 are skipped. Black and git diff --check pass. Prepared with Codex assistance.

